PR #1, opened 2016-07-11
Fixed windows path bug

caponica:
Updated driver detection to work with Windows file paths, which have "\" instead of "/".


pjedrzejewski:
Hi Christian! This repo is a read-only subtree split - could you open this PR against main repository? Sylius/Sylius. Thanks!

PR #3, opened 2019-05-14
[ORM] Refactor expresssion builder to add joins when needed

GSadee:
I'm not sure if it is a good base branch for this PR.

This change is necessary for filtering by a nested entity: https://github.com/Sylius/Sylius/pull/10382/files#diff-decadb1c0264a2aa9b1d8a2b521fbbf2R49

### DIFF
--- a/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
+++ b/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
@@ -155,7 +155,7 @@ final class ExpressionBuilder implements ExpressionBuilderInterface
     private function getFieldName(string $field): string
     {
         if (false === strpos($field, '.')) {
-            return $this->queryBuilder->getRootAliases()[0] . '.' . $field;
+            return sprintf('%s.%s', $this->getRootAlias(), $field);
         }
 
         return $field;
--- a/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
+++ b/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
@@ -27,33 +27,21 @@ final class ExpressionBuilder implements ExpressionBuilderInterface
         $this->queryBuilder = $queryBuilder;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function andX(...$expressions)
     {
         return $this->queryBuilder->expr()->andX(...$expressions);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function orX(...$expressions)
     {
         return $this->queryBuilder->expr()->orX(...$expressions);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function comparison(string $field, string $operator, $value)
     {
         return new Comparison($field, $operator, $value);
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function equals(string $field, $value)
     {
         $field = $this->adjustField($field);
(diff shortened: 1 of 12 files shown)


Zales0123:
I see three ways:

- we could make a detailed PHPSpec test for that (but it can be a pain in the neck, to be honest)
- we could create a unit test with PHPUnit with some most well-known use-cases (one field without nesting, two fields with one-level nesting, one field with two-levels nesting, etc.)
- indeed create some test application, but we must be aware that it would require a lot of additional code - it's not a plugin but a bundle, it does not have a test application already inside

I would personally give a try for the 2nd way (even though I'm not 100% sure will it work smoothly 🚀) and maybe think about other ways it fails. cc @pamil your opinion would be significant 🖖

lchrusciel:
Just for the record, there is a sample app: https://github.com/Sylius/SyliusGridBundle/tree/master/src/Bundle/test

Zales0123:
My bad 😄 🖖 Still, the PHPUnit test can be sufficient (or not, but we should check it out)

GSadee:
For now, I've added a PHPSpec test and I hope it can be merged in this form.

Additionally, I've added an issue with missing proper tests (https://github.com/Sylius/SyliusGridBundle/issues/4) and I'll try to take care of it in the nearest future.

src/Bundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -27,188 +27,135 @@ public function __construct(QueryBuilder $queryBuilder)
...
-     */
     public function equals(string $field, $value)
     {
-        $rootAlias = $this->queryBuilder->getRootAlias();
-        if (0 === strpos($field, $rootAlias . '.')) {
-            $field = substr_replace($field, '', 0, strlen($rootAlias) + 1);
-        }
-
+        $field = $this->adjustField($field);
         $parameterName = $this->getParameterName($field);
```
pamil: Are we sure that we want to run `$this->getParameterName()` with **adjusted field** as an argument? Isn't it changing the previous behaviour and breaking BC?

GSadee: `adjustField` is a method that coheres the given field, because field could look like `o.order` or `order`. So yes, the name returned by `getParameterName` could be now different (e.q. `order` instead of `o_order`), because I always remove the root alias, but I don't think that this is a BC break. The same grid configuration will work in the same way as it was before.

src/Component/Provider/ArrayGridProvider.php
```diff
@@ -57,6 +57,6 @@ private function extend(array $gridConfiguration, array $parentGridConfiguration
 
         unset($configuration['extends']);
 
-        return $configuration;
+        return $configuration !== null ? $configuration : [];
```
lchrusciel: Won't it be the same?
```suggestion
        return $configuration ?? [];
```

src/Bundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -27,164 +27,135 @@ public function __construct(QueryBuilder $queryBuilder)
...
+        $field = $this->adjustField($field);
+
+        return $this->queryBuilder->addOrderBy($this->resolveFieldByAddingJoins($field), $direction);
     }
 
     private function getFieldName(string $field): string
     {
         if (false === strpos($field, '.')) {
-            return $this->queryBuilder->getRootAlias() . '.' . $field;
+            return $this->queryBuilder->getRootAliases()[0] . '.' . $field;
```
lchrusciel:
```suggestion
            return $this->getRootAlias()()[0] . '.' . $field;
```

pamil:
Thanks, Grzegorz! :tada:

PR #5, opened 2019-07-04
Fix tests

Zales0123:


pamil:
Why is it needed?

Zales0123:
To fix tests 😄

pamil:
I'd go for dropping the problematic test instead:

- it's `StringFieldType`, so it shouldn't be expected for it to handle something that can't be cast to string
- its output is rendered directly in the grid HTML, so we need to take care about sanitizing it
- we can't sanitize the output if we just return it
- actually, if you pass something that can't be cast to a string, it'll still fail but in a different place
- even if this technically is a BC break, it'd be a really rare edge-case and we can still make it according to our BC policy (security reasons)

We can add a new test that makes sure that objects with `__toString()` also gets escaped (will be easy to do using anonymous class). What do you think?

pamil:
Fixed in #7.

PR #6, opened 2019-07-04
Fix default datetime format

Zales0123:


lchrusciel:
Thanks, Mateusz! :1st_place_medal:

PR #17, opened 2019-10-16
Add embeddable support

pamil:


src/Bundle/test/src/AppBundle/Entity/Price.php
```diff
@@ -0,0 +1,43 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace AppBundle\Entity;
+
+final class Price
+{
+    /** @var string|null */
```
GSadee: `string` -> `int` ?

src/Bundle/test/src/AppBundle/Resources/config/doctrine/Price.orm.yml
```diff
@@ -0,0 +1,9 @@
+AppBundle\Entity\Price:
+    type: embeddable
+    fields:
+        amount:
+            type: integer
+        currencyCode:
+            type: string
+            length: 255
+
```
GSadee: Unnecessary blank line

src/Bundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -195,70 +184,93 @@ private function adjustField(string $field): string
...
-                ++$key;
+            $metadata = $this->queryBuilder->getEntityManager()->getClassMetadata(
+                $metadata->getAssociationMapping($associationField)['targetEntity']
+            );
+            $rootAndAssociationField = sprintf('%s.%s', $rootField, $associationField);
 
-                continue;
-            }
+            /** @var Join[] $joins */
+            $joins = array_merge([], ...array_values($this->queryBuilder->getDQLPart('join')));
```
lchrusciel: `array_merge([], ...array_values($this->queryBuilder->getDQLPart('join')));` maybe it could be some (even private static) helper function at the end of file with meaningful name like flattenArray?

src/Bundle/Tests/Responses/Expected/american_authors_sorted_ascending.json
```diff
@@ -26,7 +26,8 @@
                         "id": "@integer@",
                         "name": "American"
                     }
-                }
+                },
+                "price": "@array@"
```
lchrusciel: Shouldn't we assert values here?

PR #21, opened 2019-10-18
Fix getting metadata for custom JOINs on not-mapped associations

### LINKED ISSUE #20: Cannot filter by joined entity since 1.7.0
Since b86d605c84be007d8b6e1533ac6533289d5626ca (cc @pamil), QB's like

```php
public function createListQueryBuilder(string $designerId): QueryBuilder
{
    return $this->createQueryBuilder('do')
        ->leftJoin(Order::class, 'parentOrder', Join::WITH, 'parentOrder.id = do.order')
        ->andWhere('do.designer = :designerId')
        ->setParameter('designerId', $designerId)
    ;
}
```

with request queries like
```php
['criteria' => ['paymentState' => ['type' => 'in', 'value' => '1,2']]]
```

and grid config like
```yml
sylius_grid:
  grids:
    app_admin_api_designer_order:
      driver:
        name: doctrine/orm
        options:
          class: "%app.model.designer_order.class%"
          repository:
            method: createListQueryBuilder
            arguments: [$designerId]
      sorting:
        createdAt: desc
      fields:
        paymentState:
          type: string
      filters:
        paymentState:
          type: string
          options:
            type: in
            fields: [parentOrder.paymentState]
```

result in exception

> Could not get mapping for "AppBundle\Entity\Order.paymentState"." at /calida/shop/vendor/[sylius/grid-bundle/src/Bundle/Doctrine/ORM/ExpressionBuilder.php line 277](https://github.com/Sylius/SyliusGridBundle/blob/b86d605c84be007d8b6e1533ac6533289d5626ca/src/Bundle/Doctrine/ORM/ExpressionBuilder.php#L277)

This is because while loop searches only across rootAliases, which are as far as I know FROM statements only.

You might think solution is to change query request from `paymentState` to `parentOrder.paymentState` (which is BC break by itself), but then this condition fails https://github.com/Sylius/SyliusGridBundle/blob/8cf0f288470ffe397dbc37bf7390ff92b72b0beb/src/Component/Filtering/FiltersApplicator.php#L48 and no filter is being applied. So then you change this key in sylius_grid config as well, but that's not enough - ExpressionBuilder still does not care about JOINed tables.

pamil:
Fixes #20.

### DIFF
--- a/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
+++ b/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\Doctrine\ORM;
 
-use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Query\Expr\Comparison;
 use Doctrine\ORM\Query\Expr\From;
 use Doctrine\ORM\Query\Expr\Join;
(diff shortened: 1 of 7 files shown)


lchrusciel:
Thank you, Kamil! :tada:

PR #24, opened 2019-10-31
fix cs issues

vvasiloi:
Ran `vendor/bin/ecs check src --fix`


Zales0123:
Thank you, Victor! :tada:

PR #28, opened 2019-12-01
[WIP] Fix SQL join with existing alias

loic425:
When you add two joins with same resource and 2 aliases, the where condition is wrong.
It's a common case when you filters your products grid with attribute values. I added a simplified example with book attributes.

DQL with the PHPunit scenario **it_filters_books_by_size_and_condition**
```
SELECT b
FROM AppBundle\Entity\Book b LEFT JOIN b.attributes size
WITH size.code = :sizeCode LEFT JOIN b.attributes condition
WITH condition.code = :conditionCode
WHERE size.value = :size_value AND size.value = :condition_value
ORDER BY b.title asc
```

This pull request doesn't fix nothing for now. It only adds the PHPunit test to expose the issue.

### DIFF
--- a/src/Bundle/Tests/DataFixtures/ORM/fixtures.yml
+++ b/src/Bundle/Tests/DataFixtures/ORM/fixtures.yml
@@ -15,15 +15,35 @@ AppBundle\Entity\Author:
         name: "<firstName()> <lastName()>"
         nationality: "@nationality_english"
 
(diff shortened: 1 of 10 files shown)


loic425:
@lchrusciel @pamil @Zales0123 
The issue was introduced with https://github.com/Sylius/SyliusGridBundle/pull/3

PR #30, opened 2020-01-02
Add support for php 7.4 and symfony 4.4

dannyvw:
Support for Symfony version 3.4 is not fixed because it will be removed in https://github.com/Sylius/SyliusGridBundle/pull/34


pamil:
Thank you, Danny! :1st_place_medal:

PR #31, opened 2020-01-02
Allow twig 3.x

dannyvw:
Deprecated class is fixed in https://github.com/Sylius/SyliusGridBundle/pull/33


pamil:
Thank you, Danny! :tada:

PR #40, opened 2020-01-03
Fix event dispatcher deprecation

dannyvw:
User Deprecated: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.


pamil:
Looks like we did already these changes in some other PR, thanks for your effort and hopefully Symfony 5 support is just around the corner :)

PR #41, opened 2020-01-03
Submit form only if criteria is available

dannyvw:
Form submit is always called, so on overviews without criteria it does unnecessary validator calls (one per filter).

### DIFF
--- a/src/Bundle/Renderer/TwigGridRenderer.php
+++ b/src/Bundle/Renderer/TwigGridRenderer.php
@@ -125,7 +125,9 @@ final class TwigGridRenderer implements GridRendererInterface
         );
 
         $criteria = $gridView->getParameters()->get('criteria', []);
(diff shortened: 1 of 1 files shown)


src/Bundle/Renderer/TwigGridRenderer.php
```diff
@@ -125,7 +125,9 @@ public function renderFilter(GridViewInterface $gridView, Filter $filter)
         );
 
         $criteria = $gridView->getParameters()->get('criteria', []);
-        $form->submit($criteria);
+        if ($criteria) {
```
lchrusciel:
```suggestion
        if ([] !== $criteria) {
```

Roshyo:
@dannyvw The code changed slightly on this matter. How do you feel about opening a new PR and adding tests about this ?

PR #44, opened 2020-01-13
Fix sorting grids by nullable path

pamil:
#### Background

- Author can have a nationality assigned, but does not have to
- Nationality have a name
- There are three authors with nationality assigned and two without

#### Actual behaviour

When I sort by nationality, it shows only the three authors that have a nationality assigned

#### Expected behaviour

When I sort by nationality, I would expect it to include all five authors, treating the ones without the nationality as it would be empty

### DIFF
--- a/src/Bundle/Tests/Functional/GridApiTest.php
+++ b/src/Bundle/Tests/Functional/GridApiTest.php
@@ -229,6 +229,23 @@ final class GridApiTest extends JsonApiTestCase
         $this->assertSame('A Study in Scarlet', $this->getFirstItemFromCurrentResponse()['title']);
     }
 
+    /** @test */
+    public function x(): void
+    {
+        $this->client->request('GET', '/authors/');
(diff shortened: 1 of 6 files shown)


src/Bundle/Tests/Functional/GridApiTest.php
```diff
@@ -229,6 +229,23 @@ public function it_filters_books_by_author_when_an_author_is_used_in_join_in_que
         $this->assertSame('A Study in Scarlet', $this->getFirstItemFromCurrentResponse()['title']);
     }
 
+    /** @test */
+    public function x(): void
```
lchrusciel: xD

pamil: x... xD

lchrusciel:
```suggestion
    public function xD(): void
```

PR #46, opened 2020-01-23
[WIP] Init make sylius grid command

loic425:
### Grid configuration
- [x] driver.name
- [x] driver.options.class
- [ ] driver.options.repository
- [x]  fields.{property}.type
- [x]  fields.{property}.label
- [x]  fields.{property}.sortable
- [x]  fields.{property}.options.template (for twig type)
- [x]  fields.{property}.options.format (for datetime type)
- [x]  sorting.{property}
- [x]  actions.main
- [x]  actions.item
- [x]  actions.bulk
- [x]  filters.type
- [x]  filters.label
- [x]  filters.options.field
- [x]  filters.options.fields
- [x]  filters.form_options
- [ ]  limits

### Import the new grid file configuration
- [ ] Import the new grid file configuration.

### Tests
- [x] Phpspec for services
- [ ] PHPUnit for MakerGridCommand

### Bundle configuration

I think we can add some configuration on the bundle such as the grid configuration path 
We can also add a naming strategy configurable.

### DIFF
--- a/src/Bundle/Maker/MakeGrid.php
+++ b/src/Bundle/Maker/MakeGrid.php
@@ -299,6 +299,7 @@ final class MakeGrid extends AbstractMaker
             $this->gridHelper->getFilterIds()
         );
         $filterData['type'] = $io->askQuestion($choiceQuestion);
+        $filterData['label'] = $io->ask('Enter the filter label', null);
 
         $choiceQuestion = new ChoiceQuestion(
             'Enter the filter options (Leave blank if you have no options',
--- a/src/Bundle/Maker/MakeGrid.php
+++ b/src/Bundle/Maker/MakeGrid.php
@@ -424,7 +424,7 @@ final class MakeGrid extends AbstractMaker
         $resource = $this->resourceHelper->getResourceNameFromAlias($resourceAlias);
         $filename = $resource.'.yaml';
 
-        return sprintf('%s/%s/%s', $this->projectDir.'/config/packages/grids', $section, $filename);
+        return sprintf('%s/%s/%s', $this->projectDir.'/config/grids', $section, $filename);
     }
 
     private function filterSortableFields(array $fields): array
--- a/src/Bundle/Maker/MakeGrid.php
+++ b/src/Bundle/Maker/MakeGrid.php
@@ -12,8 +12,8 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\Maker;
 
-use Sylius\Component\Grid\Maker\Filter\Helper\GridHelperInterface;
-use Sylius\Component\Grid\Maker\Filter\ResourceHelperInterface;
+use Sylius\Component\Grid\Maker\Helper\GridHelperInterface;
+use Sylius\Component\Grid\Maker\Helper\ResourceHelperInterface;
 use Symfony\Bundle\MakerBundle\ConsoleStyle;
 use Symfony\Bundle\MakerBundle\DependencyBuilder;
 use Symfony\Bundle\MakerBundle\Generator;
@@ -136,6 +136,19 @@ final class MakeGrid extends AbstractMaker
             $fields = array_merge($fields, $newField);
         }
 
+        $sortableFields = $this->filterSortableFields($fields);
+        $sortingFieldNames = [];
+
+        if (count($sortableFields) > 0) {
+            $sortingFieldNames = $this->askForSortingFieldNames($io, $sortableFields);
+        }
+
+        $sortingFields = [];
+        foreach ($sortingFieldNames as $fieldName) {
+            $order = $this->askForSortingOrder($io, $fieldName);
+            $sortingFields[$fieldName] = $order;
+        }
+
         $actions = [
             'main' => [],
             'item' => [],
@@ -153,7 +166,7 @@ final class MakeGrid extends AbstractMaker
             }
         }
 
-        $this->generateGridConfigFile($input, $io, $generator, $fields, $actions);
+        $this->generateGridConfigFile($input, $io, $generator, $fields, $sortingFields, $actions);
     }
 
     private function askForNextField(ConsoleStyle $io, bool $isFirstField)
@@ -167,12 +180,7 @@ final class MakeGrid extends AbstractMaker
         }
 
         $fieldName = $io->ask($questionText, null, function ($name): ?string {
-            // allow it to be empty
-            if (!$name) {
-                return $name;
-            }
-
-            return $name;
+            return $name ?: null;
         });
 
         if (!$fieldName) {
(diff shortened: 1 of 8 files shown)


src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Yaml\Yaml;
+use Webmozart\Assert\Assert;
+
+final class MakeGrid extends AbstractMaker
+{
+    public static $defaultActionTypes = [
```
lchrusciel:
```suggestion
    private static $defaultActionTypes = [
```

Also, won't it be better to make this variable a constant?  In case we would like to change it, it should be injected.

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+    {
+        $io->writeln('');
+
+        if ($isFirstField) {
+            $questionText = 'New field name (press <return> to stop adding fields)';
+        } else {
+            $questionText = 'Add another field? Enter the field name (or press <return> to stop adding fields)';
+        }
+
+        $fieldName = $io->ask($questionText, null, function ($name) {
```
lchrusciel:
```suggestion
        $fieldName = $io->ask($questionText, null, function ($name): ?string {
```

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+            // allow it to be empty
+            if (!$name) {
+                return $name;
+            }
+
+            return $name;
```
lchrusciel:
```suggestion
            return $name ?: null;
```
and removal of if statement above?

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+        });
+
+        if (!$fieldName) {
+            return null;
+        }
+
+        $fieldData = [];
+        $fieldData['type'] = $io->choice('Enter the field type', ['string', 'datetime', 'twig'], 'string');
```
lchrusciel: Later, we should probably take it from `sylius.registry.grid_field` parameter

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+        InputInterface $input,
+        ConsoleStyle $io,
+        Generator $generator,
+        array $fields,
+        array $actions
+    ): void {
+        $section = $input->getArgument('section');
+        $resourceAlias = $input->getArgument('resource');
+        [$appName, $resourceName] = $this->resourceHelper->splitResourceAlias($resourceAlias);
+        $gridId = sprintf('%s_%s_%s', $appName, $section, $resourceName);
```
lchrusciel: In the long run, this should be configurable as well, I suppose

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,285 @@
...
+
+        $this->fileSystem->dumpFile($gridConfigDir, $yaml);
+    }
+
+    private function getGridConfigDir(string $resourceAlias, string $section)
+    {
+        $resource = $this->resourceHelper->getResourceNameFromAlias($resourceAlias);
+        $filename = $resource.'.yaml';
+
+        return sprintf('%s/%s/%s', $this->projectDir.'/config/packages/grids', $section, $filename);
```
lchrusciel: I'm not sure about adding `grid` subfolder to packages, as I found it confusing with `env` folders

src/Component/Maker/Helper/GridHelper.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Maker\Helper;
+
+class GridHelper implements GridHelperInterface
```
lchrusciel:
```suggestion
final class GridHelper implements GridHelperInterface
```

src/Component/Maker/Helper/ResourceHelper.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Maker\Helper;
+
+class ResourceHelper implements ResourceHelperInterface
```
lchrusciel:
```suggestion
final class ResourceHelper implements ResourceHelperInterface
```

lchrusciel: Shouldn't we mark these classes as `@internal`? /cc @pamil

src/Component/spec/Maker/Helper/GridHelperSpec.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace spec\Sylius\Component\Grid\Maker\Helper;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Grid\Maker\Helper\GridHelper;
+
+class GridHelperSpec extends ObjectBehavior
```
lchrusciel:
```suggestion
final class GridHelperSpec extends ObjectBehavior
```

src/Component/spec/Maker/Helper/GridHelperSpec.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    function let(): void
+    {
+        $this->beConstructedWith([
+            'string' => "@SyliusUi/Grid/Filter/string.html.twig",
+            'boolean' => "@SyliusUi/Grid/Filter/boolean.html.twig",
+        ]);
+    }
+
+    function it_is_initializable()
```
lchrusciel:
```suggestion
    function it_is_initializable(): void
```
or let's remove it. Same for the rest of the specs

src/Bundle/Maker/MakeGrid.php
loic425: Yes thank you, I was searching for a parameter like that :)

src/Bundle/Maker/MakeGrid.php
loic425: I think this path should be configurable on the bundle.
We can also ask for it on command when it's not configured on the bundle configuration.

src/Bundle/Maker/MakeGrid.php
loic425: Good idea.

src/Bundle/Maker/MakeGrid.php
loic425: public/protected/private const?

src/Bundle/Maker/MakeGrid.php
loic425: But you're right, by default, I should move them to config/grids directory (On Monofony project too).

src/Bundle/Maker/MakeGrid.php
lchrusciel: We should be as strict as possible. We will always be able to expose const, but we won't be able to reduce the scope. `private const`

PR #49, opened 2020-02-19
Add multiple values support in SelectFilter

### LINKED ISSUE #48: Unable to use Select filter with multiple values
I have configured a new grid for order with state filter. I want filter order with state in list.

Config:

```yaml
sylius_grid:
    grids:
        sylius_admin_order:
            filters:
                state:
                    type: select
                    form_options:
                        multiple: true
                        choices:
                            cart: cart
                            new: new
                            cancelled: cancelled
                            fulfilled: fulfilled
```

Result:

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; 
check the manual that corresponds to your MySQL server version for the right syntax to use near ' 
'fulfilled' ORDER BY s0_.number DESC LIMIT 10' at line 1
```

macintoshplus:
| Q               | A
| --------------- | -----
| Branch?         | branch 1.6 does not exist, master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #48
| License         | MIT

### DIFF
--- a/src/Bundle/Tests/Functional/GridApiTest.php
+++ b/src/Bundle/Tests/Functional/GridApiTest.php
@@ -234,9 +234,9 @@ final class GridApiTest extends JsonApiTestCase
     {
         $this->client->request('GET', '/books/?criteria[state][]=unpublished&criteria[state][]=initial');
 
-        $this->assertCount(2, $this->getItemsFromCurrentResponse());
+        $this->assertCount(2, $result = $this->getItemsFromCurrentResponse());
         $this->assertSame('A Study in Scarlet', $this->getFirstItemFromCurrentResponse()['title']);
-        $this->assertSame('The Lost World', end($this->getItemsFromCurrentResponse())['title']);
+        $this->assertSame('The Lost World', end($result)['title']);
     }
 
     private function getItemsFromCurrentResponse(): array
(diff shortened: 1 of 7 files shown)


macintoshplus:
Thanks for your return.

I have added a new field 'state' with a filter on the Book entity.

src/Bundle/Tests/Functional/GridApiTest.php
```diff
@@ -229,6 +229,16 @@ public function it_filters_books_by_author_when_an_author_is_used_in_join_in_que
...
     }
 
+    /** @test */
+    public function it_filters_books_by_multiple_state(): void
+    {
+        $this->client->request('GET', '/books/?criteria[state][]=unpublished&criteria[state][]=initial');
+
+        $this->assertCount(2, $result = $this->getItemsFromCurrentResponse());
+        $this->assertSame('A Study in Scarlet', $this->getFirstItemFromCurrentResponse()['title']);
+        $this->assertSame('The Lost World', end($result)['title']);
```
lchrusciel: We could use some helper method here as well
```suggestion
        $this->assertSame('The Lost World', $this->getLastItemFromCurrentResponse()['title']);
```

macintoshplus: This helper does not exist. Do you want I write this helper for one use?

lchrusciel: I meant the creation of the new one ;) But, don't worry, it's good to go for me. Probably after Sylius 1.7 release

macintoshplus: Ok, this bug fix will be apply only for the version 1.7 or older too ?

macintoshplus:
Ok, this bug fix will be apply only for the version 1.7 or older too ?

lchrusciel:
Grid bundle has a separate release cycle, so should be available despite sylius release

lchrusciel:
Thanks, Jean-Baptiste! :tada:

Issue #42, opened 2020-01-07
Version 1.7.4 has problem with sorting grids by names


Tomanhez:
Sorting grids by names doesn't work properly

For example : In Sylius/Sylius we have broken behat test features/payment/managing_payment_methods/sorting_payment_methods.feature

![Screenshot 2020-01-07 at 15 42 21](https://user-images.githubusercontent.com/39232096/71903411-532a6780-3164-11ea-9709-9636cc6b1783.png)

F-Locos:
Hello,

I'm confirming this issue and not only on names. I got it on the products taxons page on the position field.
I also found some weird behaviour like : 
- Wrong amount of products
- A same product on different pages (should not)

Tomanhez:
Hi @F-Locos 

Try use `composer update` , few moments ago we added [hot-fix](https://github.com/Sylius/Sylius/pull/11005) that downgrade GridBundle from 1.7.4 to 1.7.3 on branches from 1.5 to latest,  Now we work on fixing 1.7.4

GSadee:
I guess it is fixed by merging #55

PR #55, opened 2020-04-14
Revert "Fix autojoining with multiple aliases"

GSadee:
We need to revert the changes from Sylius/SyliusGridBundle#35 because they cause the other issue (https://github.com/Sylius/SyliusGridBundle/issues/42), but we definitely need to solve all these problems correctly.


lchrusciel:
Thank you, Grzegorz! :tada:

PR #64, opened 2020-08-19
[WIP] Allow falsy values

### LINKED ISSUE #63: Broken Select filter if filtering by falsy values
The Select filter doesn't work for falsey values (like string '0' or simply false), because:

        if (empty($data)) {
            return;
        }

mikemix:
Fixes #63


lchrusciel:
Hey Mike,

thanks for your PR, however, the build is red and new logic is not tested. Without it, we cannot merge it.

mikemix:
I'm aware of that, just had start doing my usual business instead. For the moment I've changed the title to indicate the WIP status, thanks.

PR #65, opened 2020-08-19
Update code to support Symfony 5

### LINKED ISSUE #29: Symfony 5 support
The issue to track and coordinate efforts.

stloyd:
Related to #29, it will not fully fix it due to lock on `sylius/currency-bundle`, see #66 for possible fix.


stloyd:
Friendly ping @pamil 🥰

lchrusciel (approved):

src/Component/Event/GridDefinitionConverterEvent.php
```diff
@@ -14,7 +14,7 @@
 namespace Sylius\Component\Grid\Event;
 
 use Sylius\Component\Grid\Definition\Grid;
-use Symfony\Component\EventDispatcher\Event;
+use Symfony\Contracts\EventDispatcher\Event;
```
lchrusciel, line 17: To support both versions of Symfony with BC we've created a polyfil at SyliusLabs. Can you add https://github.com/SyliusLabs/PolyfillSymfonyEventDispatcher here? For reference you may check: https://github.com/Sylius/SyliusMailerBundle/commit/e3f1042caf1e7f408e1b992742630124be30fa93

stloyd: @lchrusciel Sorry, but I'm not fully sure, why we should use this polyfill. Symfony already has some kind of BC, and version 4.4 is already forcing to use `Contracts` namespace:
https://github.com/symfony/symfony/blob/4.4/src/Symfony/Contracts/EventDispatcher/Event.php

lchrusciel: The build-in polyfill is not working with us. If one would use our current event system, they may have typehint in some event listeners to `Symfony\Component\EventDispatcher\Event`, which is perfectly valid. If we change the class we are extending from here, all these listeners will stop working. We are allowed to switch namespace only for Symfony 5, as it is not possible to extend `Symfony\Component\EventDispatcher\Event` anymore. We have to use component namespace as long as we can due to BC Policy. Therefore our polyfill supports extending of deprecated code from component namespace, but once this class is not longer available, we will switch to contracts namespace.

stloyd:
@lchrusciel @pamil Is there anything missing?

pamil:
Thanks a lot, most of these changes has already happened through some other PRs in the meanwhile and official support in #86.

PR #67, opened 2020-10-11
Inject repository through the container instead of doctrine

dnna:
When using this bundle as part of a Sylius project that also uses ResourceBundle, the repository instance that gets used for queries is not the same instance as the one defined by the ResourceBundle. The reason is that the repository here is injected directly by Doctrine, skipping the relevant service definitions set in the container.

This causes problems when you try to inject dependencies into the repository instance, because even though the service configuration in ResouceBundle works fine, when a function like `createListQueryBuilder` is called by GridBundle, the injected dependencies will be null.

This PR attempts to solve this by first looking for a service definition in ResourceBundle and if so using that, otherwise falling back to the original functionality. This solution is not very elegant (we have to inject the whole contianer) and there might a better way of doing, but I would like to at least open the discussion of how this could be done.

### DIFF
--- a/src/Bundle/Doctrine/ORM/Driver.php
+++ b/src/Bundle/Doctrine/ORM/Driver.php
@@ -13,9 +13,12 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\Doctrine\ORM;
 
+use Doctrine\Common\Persistence\ManagerRegistry;
+use Doctrine\Common\Persistence\ObjectManager;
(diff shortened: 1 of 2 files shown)


lchrusciel:
Related to https://github.com/Sylius/SyliusResourceBundle/pull/218

pamil:
Thanks for the PR, the underlying issue is fixed in Sylius/SyliusResourceBundle#218 and this one should not be needed then :)

PR #68, opened 2020-10-18
Add possibility to set pagination options on grids

### LINKED ISSUE #10343: Product pagination bug (again)
**Sylius version affected**: 1.4.1 and up (didn't test lower versions)
**Description**  
The same bug described here: https://github.com/Sylius/Sylius/issues/9699
Opened this issue to keep record.

**Steps to reproduce**  
It's not persistent across all categories.
Here is one where the bug occurs: https://demo.sylius.com/en_US/taxons/t-shirts?limit=9

**Possible Solution**  
https://github.com/Sylius/Sylius/issues/9699#issuecomment-487631905

vvasiloi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/10343
| License         | MIT

A similar solution can be applied to the Resource Bundle https://github.com/Sylius/SyliusResourceBundle/issues/148.

The solution is based on this experiment: https://github.com/vvasiloi/doctrine-pagination

TODO:
- [x] test setting default values by parameters
- [x] add parameters for setting default values
- [x] test setting the values per grid
- [x] allow setting the values per grid
- [x] test with fetch join enabled/disabled per grid
- [x] test with fetch join enabled/disabled by default
- [x] test with order by to-many association field and allow output walkers enabled/disabled per grid
- [x] test with order by to-many association field and allow output walkers enabled/disabled by default
- [x] update documentation

Additional:
- [ ] apply a similar solution to Sylius Resource Bundle
- [ ] enable fetch join and output walkers in Sylius Standard 
- [ ] use configuration instead of parameters?


vvasiloi:
@lchrusciel can I have a review, please? I need to know if and how I should move forward with this.

src/Bundle/Doctrine/ORM/DataSource.php
```diff
@@ -65,7 +83,7 @@ public function getExpressionBuilder(): ExpressionBuilderInterface
     public function getData(Parameters $parameters)
     {
         // Use output walkers option in DoctrineORMAdapter should be false as it affects performance greatly. (see #3775)
```
lchrusciel, line 85: I would move this comment to construct docblock & probably we should write a short doc about it

lchrusciel (commented):
Hey Victor,

this issue is with us for way too long. Your proposal seems to be in a good direction. However generic setting could be done with configuration component indeed. I would like to see it merged in SyliusGridBundle. 

/cc @pamil, what is your opinion?

pamil (approved):
Thanks a lot and sorry for the delay! I've just rebased the PR to the newest master so that CI does not fail anymore. Also resolved @lchrusciel comment and added a comment not to rely on these parameters for defining global behaviour for now.

pamil:
I think in the current form it's pretty good to go. I'd say the next steps, for now, would be "test with order by to-many association field and allow output walkers enabled/disabled per grid" and "update documentation".

Then it should be good to merge, we can skip configurable default settings for all grids for now and address that later. If so, I think the best place to set it would be `sylius_grid.drivers.orm` configuration.

Also, really cool work on https://github.com/vvasiloi/doctrine-pagination! I wonder if there's some solution that is both performant and accurate? Can it be solved somehow with a custom query (for cases like Sylius/Sylius#3775, when using output walkers takes too much time)?

vvasiloi:
@AdamKasp Thanks for picking it up, I forgot about it... 😞 Do you need anything from me?

vvasiloi:
The "use output walker" parameters should allow `null`, because Doctrine allows it to be null, in which case it will use query hints to determine if an output walker should be used.

src/Bundle/Doctrine/ORM/Driver.php
```diff
@@ -43,8 +43,11 @@ public function getDataSource(array $configuration, Parameters $parameters): Dat
         /** @var EntityRepository $repository */
         $repository = $manager->getRepository($configuration['class']);
 
+        $fetchJoinCollection = $configuration['pagination']['fetch_join_collection'] ?? true;
+        $useOutputWalkers = $configuration['pagination']['use_output_walkers'] ?? true;
+
         if (!isset($configuration['repository']['method'])) {
-            return new DataSource($repository->createQueryBuilder('o'));
+            return new DataSource($repository->createQueryBuilder('o'), $fetchJoinCollection, $useOutputWalkers);
```
lchrusciel, line 50: Just a minor thought, but perhaps it would be better to pass the whole configuration array into the data source?

lchrusciel (approved):

src/Bundle/Doctrine/ORM/Driver.php
GSadee: I have no strong opinion about that change. But now, it looks that every operation on the configuration array occurs in Driver, so it makes sense to do it there and pass only proper parameters to DataSource. Do you see any benefits or potential benefits of changing it?

lchrusciel (approved):

lchrusciel:
Thank you, Victor! :tada:

vvasiloi:
Thank you, Adam and Grzegorz!
I will try to not abandon my PRs like that anymore.

PR #120, opened 2021-02-05
Update default limits to more sensible values

loevgaard:
I have been annoyed by these default limits multiple times - maybe others have as well? :-)


Roshyo:
What do you think to have rather a config for defining all grids at once ?

Like: 
```yaml
sylius_grid:
    default_grids_limits: [50, 100, 200]
```

And then the individual grid definition would override this.
The thing is that the limits you consider as "more sensible" may not be the same for everyone.

PR #146, opened 2021-03-29
Added NumericRange filter which gives the possibility of filtering …

christopherhero:
| Q               | A
| --------------- | -----
| Branch?         | 1.13
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #
| License         | MIT

While working on adding a custom filter, I thought that it could be in the bundle.
I add `range numerical` filter which can filter on numerical fields in range (from-to) Like existing DateFilter.

Also in this pull request is a small change in two classes:
- `Doctrine\ORM\ExpressionBuilder`
- `Doctrine\ORM\DataSource`.

I suggested adding filtering capability also when using HAVING instead of WHERE in the query. For example, when we use `count (*)`. In current architecture is hard to distinguish the type of condition (WHERE or HAVING). Another way than suggested by me is to add pubilic methods e.g `havingLessThan` to `ORM\ExpressionBuilder` and new interface "HavingExpressionBuilder"

Please give your opinion what do you think about it?

This idea also came while working on this filter because I had to filter in queries with the `HAVING` clause.

### DIFF
--- a/src/Bundle/Resources/config/services/filters.xml
+++ b/src/Bundle/Resources/config/services/filters.xml
@@ -67,6 +67,16 @@
         </service>
         <service id="sylius.form.type.grid_filter.exists" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\ExistsFilterType" />
 
+        <service id="Sylius\Component\Grid\Filter\NumericRangeFilter">
+            <tag name="sylius.grid_filter" type="numeric_range" form-type="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType" />
+        </service>
+        <service id="sylius.grid_filter.numeric_range" alias="Sylius\Component\Grid\Filter\NumericRangeFilter" />
+
+        <service id="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType">
+            <tag name="form.type" />
+        </service>
+        <service id="sylius.form.type.grid_filter.numeric_range" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType" />
+
         <service id="Sylius\Component\Grid\Filter\SelectFilter">
             <tag name="sylius.grid_filter" type="select" form-type="Sylius\Bundle\GridBundle\Form\Type\Filter\SelectFilterType" />
         </service>
--- a/src/Component/Filter/NumericRangeFilter.php
+++ b/src/Component/Filter/NumericRangeFilter.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Filter;
+
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class NumericRangeFilter implements FilterInterface
+{
+    public const DEFAULT_SCALE = 0;
+
+    public const DEFAULT_ROUNDING_MODE = \NumberFormatter::ROUND_HALFUP;
+
+    public const DEFAULT_INCLUSIVE_FROM = true;
+
+    public const DEFAULT_INCLUSIVE_TO = true;
+
+    public function apply(DataSourceInterface $dataSource, string $name, $data, array $options): void
+    {
+        if (empty($data)) {
+            return;
+        }
+
+        $field = (string) ($options['field'] ?? $name);
+        $scale = (int) ($options['scale'] ?? self::DEFAULT_SCALE);
+        $mode = (int) ($options['rounding_mode'] ?? self::DEFAULT_ROUNDING_MODE);
+
+        $greaterThan = $this->getDataValue($data, 'greaterThan');
+        $lessThan = $this->getDataValue($data, 'lessThan');
+
+        $expressionBuilder = $dataSource->getExpressionBuilder();
+
+        if ('' !== $greaterThan) {
+            $inclusive = (bool) ($options['inclusive_from'] ?? self::DEFAULT_INCLUSIVE_FROM);
+            $amount = $this->normalizeAmount((float) $greaterThan, $scale, $mode);
+
+            if ($inclusive) {
+                $dataSource->restrict($expressionBuilder->greaterThanOrEqual($field, $amount));
+            } else {
+                $dataSource->restrict($expressionBuilder->greaterThan($field, $amount));
+            }
+        }
+
+        if ('' !== $lessThan) {
+            $inclusive = (bool) ($options['inclusive_to'] ?? self::DEFAULT_INCLUSIVE_TO);
+            $amount = $this->normalizeAmount((float) $lessThan, $scale, $mode);
+
+            if ($inclusive) {
+                $dataSource->restrict($expressionBuilder->lessThanOrEqual($field, $amount));
+            } else {
+                $dataSource->restrict($expressionBuilder->lessThan($field, $amount));
+            }
+        }
+    }
+
+    private function normalizeAmount(float $amount, int $scale, int $mode): int
+    {
+        return (int) round($amount * (10 ** $scale), $mode);
+    }
+
+    /** @param array<array-key, string> $data */
+    private function getDataValue(array $data, string $key): string
+    {
+        return $data[$key] ?? '';
+    }
+}
--- a/src/Component/spec/Filter/NumericRangeFilterSpec.php
+++ b/src/Component/spec/Filter/NumericRangeFilterSpec.php
@@ -1,0 +1,229 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Component\Grid\Filter;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Data\ExpressionBuilderInterface;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class NumericRangeFilterSpec extends ObjectBehavior
+{
+    function it_implements_a_filter_interface(): void
+    {
+        $this->shouldImplement(FilterInterface::class);
+    }
+
+    function it_does_nothing_when_there_is_no_data(DataSourceInterface $dataSource): void
+    {
+        $dataSource->restrict(Argument::any())->shouldNotBeCalled();
+
+        $this->apply(
+            $dataSource,
+            'number',
+            [],
+            [],
+        );
+    }
+
+    function it_filters_number_from(
+        DataSourceInterface $dataSource,
+        ExpressionBuilderInterface $expressionBuilder,
+    ): void {
+        $expressionBuilder->greaterThanOrEqual('number', 3)->willReturn('EXPR');
+        $dataSource->getExpressionBuilder()->willReturn($expressionBuilder);
+
+        $dataSource->getExpressionBuilder()->shouldBeCalledOnce();
+        $dataSource->restrict('EXPR')->shouldBeCalledOnce();
+
+        $this->apply(
+            $dataSource,
+            'number',
+            [
+                'greaterThan' => '3',
+            ],
+            [],
+        );
+    }
+
+    function it_filters_number_from_without_inclusive_from(
+        DataSourceInterface $dataSource,
+        ExpressionBuilderInterface $expressionBuilder,
+    ): void {
+        $expressionBuilder->greaterThan('number', 7)->willReturn('EXPR');
+        $dataSource->getExpressionBuilder()->willReturn($expressionBuilder);
+
+        $dataSource->getExpressionBuilder()->shouldBeCalledOnce();
+        $dataSource->restrict('EXPR')->shouldBeCalledOnce();
+
+        $this->apply(
+            $dataSource,
+            'number',
+            [
+                'greaterThan' => '7',
+            ],
+            [
+                'inclusive_from' => false,
+            ],
+        );
+    }
+
+    function it_filters_number_to(
+        DataSourceInterface $dataSource,
+        ExpressionBuilderInterface $expressionBuilder,
+    ): void {
+        $dataSource->getExpressionBuilder()->willReturn($expressionBuilder);
+        $expressionBuilder->lessThanOrEqual('number', 8)->willReturn('EXPR');
+
+        $dataSource->getExpressionBuilder()->shouldBeCalledOnce();
+        $dataSource->restrict('EXPR')->shouldBeCalled();
+
+        $this->apply(
+            $dataSource,
+            'number',
+            [
+                'lessThan' => '8',
+            ],
+            [],
+        );
+    }
+
+    function it_filters_number_to_without_inclusive_to(
+        DataSourceInterface $dataSource,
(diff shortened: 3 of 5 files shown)


Roshyo:
Hi @christopherhero and thank you for your PR.

But I feel this one is a bit too complex for the value it should bring.
I would also love to have a numeric filter out of the box. But IMO it should just look like the current `MoneyFilter` without the currency part. What do you think ?

christopherhero:
@Roshyo Hi thanks :)


I have lightly changed `RangeNumericalFiolter` to `NumericalFIlter` (I also removed the logic for the expression builder, I will add that in a separate pr)

christopherhero:
@Roshyo what you think now :) ?

Roshyo:
LGTM. Let's wait for another approval before merging

Roshyo:
@christopherhero Could you try to rebase on 1.13?

christopherhero:
@Roshyo updated

Roshyo (approved):

src/Component/Filter/NumericalFilter.php
```diff
@@ -0,0 +1,57 @@
...
+            $dataSource->restrict($expressionBuilder->greaterThan($field, $this->normalizeAmount((float) $greaterThan, $scale)));
+        }
+        if ('' !== $lessThan) {
+            $dataSource->restrict($expressionBuilder->lessThan($field, $this->normalizeAmount((float) $lessThan, $scale)));
+        }
+    }
+
+    private function normalizeAmount(float $amount, int $scale): int
+    {
+        return (int) round($amount * (10 ** $scale));
```
vvasiloi: Let's say I want to get all the variants that have on_hold > 5, do I have to write 0.05 in the input?

christopherhero: No, in this case you have to configure `scale: 0` in your grid config e.g

```
sylius_grid:
    grids:
        ...
            filters:
                numberField:
                    type: numerical
                    form_options:
                        scale: 0
```

diimpp:
+1 for adding numeric filter, current list looks lonely without it :smiley: 

I've a naming concern, it feels more like `RangeFilter`, rather than `NumericFilter`. 

How about to synchronize naming and capabilities with api-platform?
https://api-platform.com/docs/core/filters/#numeric-filter `<eq>`
https://api-platform.com/docs/core/filters/#range-filter `<lt|gt|lte|gte|between>`

src/Component/Filter/NumericalFilter.php
vvasiloi: Cool, thanks! Maybe make 0 the default then?

christopherhero:
@diimpp Thanks for review :) 
Regarding Api platfrom, could you give some example of what you mean ? 
Because I don't fully understand, thanks :)

christopherhero:
@diimpp What else to change :) ?

src/Bundle/Form/Type/Filter/RangeNumericFilterType.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Sylius\Bundle\GridBundle\Form\Type\Filter;
+
+use Sylius\Component\Grid\Filter\RangeNumericFilter;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\NumberType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class RangeNumericFilterType extends AbstractType
```
diimpp: Either
```suggestion
final class NumericRangeFilterType extends AbstractType
```
or

```suggestion
final class RangeFilterType extends AbstractType
```

@loic425 Hi, wdyt?

src/Bundle/Form/Type/Filter/RangeNumericFilterType.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder
+            ->add('greaterThan', NumberType::class, [
+                'label' => 'sylius.ui.greater_than',
+                'required' => false,
+                'scale' => $options['scale'],
```
diimpp: Scale depends on rounding mode, so it should be possible to configure as well.

image

https://symfony.com/doc/current/reference/forms/types/number.html#rounding-mode

src/Bundle/Form/Type/Filter/RangeNumericFilterType.php
```diff
@@ -0,0 +1,55 @@
...
+            ])
+            ->add('lessThan', NumberType::class, [
+                'label' => 'sylius.ui.less_than',
+                'required' => false,
+                'scale' => $options['scale'],
```
diimpp: Same

src/Bundle/Resources/config/services/filters.xml
```diff
@@ -67,6 +67,16 @@
         </service>
         <service id="sylius.form.type.grid_filter.exists" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\ExistsFilterType" />
 
+        <service id="Sylius\Component\Grid\Filter\RangeNumericFilter">
+            <tag name="sylius.grid_filter" type="numeric" form-type="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeNumericFilterType" />
```
diimpp:
```suggestion
            <tag name="sylius.grid_filter" type="numeric_range" form-type="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType" />
```
```suggestion
            <tag name="sylius.grid_filter" type="range" form-type="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeFilterType" />
```

src/Bundle/Form/Type/Filter/RangeNumericFilterType.php
```diff
@@ -0,0 +1,55 @@
...
+                'data_class' => null,
+                'scale' => RangeNumericFilter::DEFAULT_SCALE,
+            ])
+            ->setAllowedTypes('scale', ['string', 'int'])
+        ;
+    }
+
+    public function getBlockPrefix(): string
+    {
+        return 'sylius_grid_filter_range_numeric';
```
diimpp:
```suggestion
        return 'sylius_grid_filter_numeric_range';
```
```suggestion
        return 'sylius_grid_filter_range';
```

src/Bundle/Resources/config/services/filters.xml
```diff
@@ -67,6 +67,16 @@
...
+        </service>
+        <service id="sylius.grid_filter.numeric" alias="Sylius\Component\Grid\Filter\RangeNumericFilter" />
```
diimpp:
```suggestion
        <service id="sylius.grid_filter.numeric_range" alias="Sylius\Component\Grid\Filter\NumericRangeFilter" />
```
```suggestion
        <service id="sylius.grid_filter.range" alias="Sylius\Component\Grid\Filter\RangeFilter" />
```

src/Bundle/Resources/config/services/filters.xml
```diff
@@ -67,6 +67,16 @@
...
+
+        <service id="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeNumericFilterType">
```
diimpp:
```suggestion
        <service id="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType">
```
```suggestion
        <service id="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeFilterType">
```

src/Bundle/Resources/config/services/filters.xml
```diff
@@ -67,6 +67,16 @@
...
+            <tag name="form.type" />
+        </service>
+        <service id="sylius.form.type.grid_filter.numeric" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeNumericFilterType" />
```
diimpp:
```suggestion
        <service id="sylius.form.type.grid_filter.numeric_range" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\NumericRangeFilterType" />
```
```suggestion
        <service id="sylius.form.type.grid_filter.range" alias="Sylius\Bundle\GridBundle\Form\Type\Filter\RangeFilterType" />
```

src/Component/Filter/RangeNumericFilter.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Filter;
+
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class RangeNumericFilter implements FilterInterface
```
diimpp:
```suggestion
final class NumericRangeFilter implements FilterInterface
```
```suggestion
final class RangeFilter implements FilterInterface
```

src/Component/Filter/RangeNumericFilter.php
```diff
@@ -0,0 +1,57 @@
...
+{
+    public const DEFAULT_SCALE = 0;
+
+    public function apply(DataSourceInterface $dataSource, string $name, $data, array $options): void
+    {
+        if (empty($data)) {
+            return;
+        }
+
+        $field = $options['field'] ?? $name;
```
diimpp: I assume if I will write `field: 1`, then it would be int and we need string. `Date` filter do this conversion as well.
```suggestion
        $field = (string) $options['field'] ?? $name;
```

src/Component/Filter/RangeNumericFilter.php
```diff
@@ -0,0 +1,57 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Filter;
+
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class RangeNumericFilter implements FilterInterface
+{
+    public const DEFAULT_SCALE = 0;
```
diimpp: I wonder which default scale would be sensible.

@Sylius/development-team wdyt?

SF numeric form type uses `3`.

src/Component/Filter/RangeNumericFilter.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    private function normalizeAmount(float $amount, int $scale): int
+    {
+        return (int) round($amount * (10 ** $scale));
+    }
+
+    /**
+     * @param string[] $data
+     */
```
diimpp: Since phpstan now understands more complex docblocks, something like this should be done.

```suggestion
    /** @param array<array-key, string> $data */
```

diimpp (commented):
Hi @christopherhero, sorry for late reply. :pray: 

Please see the comments.

Filter should allow to specify comparison strictness, like `lt|gt|lte|gte`, e.g. `greater than` and `greater than or equel`

christopherhero:
@diimpp

christopherhero:
what do you think about it now ;) 
@diimpp @vvasiloi

src/Component/spec/Filter/NumericRangeFilterSpec.php
```diff
@@ -0,0 +1,198 @@
...
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class NumericRangeFilterSpec extends ObjectBehavior
+{
+    function it_implements_a_filter_interface(): void
+    {
+        $this->shouldImplement(FilterInterface::class);
+    }
+
+    function it_does_nothing_when_there_is_no_data(DataSourceInterface $dataSource): void
```
vvasiloi, line 29: This should check that `$dataSource->restrict()` was not called, like in the ExistsFilter for example: https://github.com/Sylius/SyliusGridBundle/blob/5d3f702947f17502e0d6e173276086c1d8ebcf9b/src/Component/spec/Filter/ExistsFilterSpec.php#L30-L35

For some reason, the same was done for the MoneyFilter, but it does not seem right:
https://github.com/Sylius/SyliusGridBundle/blob/5d3f702947f17502e0d6e173276086c1d8ebcf9b/src/Component/spec/Filter/MoneyFilterSpec.php#L34-L42

/cc @Sylius/core-team

src/Component/spec/Filter/NumericRangeFilterSpec.php
```diff
@@ -0,0 +1,198 @@
...
+            [],
+            [],
+        );
+    }
+
+    function it_filters_number_from(
+        DataSourceInterface $dataSource,
+        ExpressionBuilderInterface $expressionBuilder,
+    ): void {
+        $expressionBuilder->greaterThanOrEqual('number', '3')->willReturn('EXPR');
```
vvasiloi: Does this work? The second argument should be an integer. 🤔

src/Component/spec/Filter/NumericRangeFilterSpec.php
```diff
@@ -0,0 +1,198 @@
...
+
+        $this->apply(
+            $dataSource,
+            'number',
+            [
+                'greaterThan' => '120.78',
+                'lessThan' => '258.51',
+            ],
+            [
+                'scale' => 0,
```
vvasiloi, line 168: Maybe include a test where the scale is not 0, since that's the default value now.

vvasiloi (changes requested):

src/Component/spec/Filter/NumericRangeFilterSpec.php
christopherhero: example -> `vendor/sylius/grid-bundle/src/Bundle/spec/Doctrine/PHPCRODM/ExpressionBuilderSpec.php:78`

christopherhero: Yes, works (but I changed)

christopherhero:
@vvasiloi

src/Component/Filter/NumericRangeFilter.php
```diff
@@ -0,0 +1,78 @@
...
+        $scale = (int) ($options['scale'] ?? self::DEFAULT_SCALE);
+        $mode = (int) ($options['rounding_mode'] ?? self::DEFAULT_ROUNDING_MODE);
+
+        $greaterThan = $this->getDataValue($data, 'greaterThan');
+        $lessThan = $this->getDataValue($data, 'lessThan');
+
+        $expressionBuilder = $dataSource->getExpressionBuilder();
+
+        if ('' !== $greaterThan) {
+            $inclusive = (bool) ($options['inclusive_from'] ?? self::DEFAULT_INCLUSIVE_FROM);
```
vvasiloi, line 45: This statement can be moved up, outside the if statement, and the one on line 57 can be removed.

src/Component/Filter/NumericRangeFilter.php
```diff
@@ -0,0 +1,78 @@
...
+            $amount = $this->normalizeAmount((float) $greaterThan, $scale, $mode);
+
+            if (true === $inclusive) {
```
vvasiloi: Inclusive is always boolean, so this can be simplified to `if ($inclusive)`. Same for the other one at line 60.

vvasiloi (approved):
Only 2 more observations I missed in the previous review, but it can be merged even without addressing those.

src/Component/Filter/RangeNumericFilter.php
vvasiloi: I think 0 is the most sensible. I don't think filtering with decimal precision is more common than with whole numbers.

src/Component/Filter/NumericRangeFilter.php
christopherhero: Unfortunately, but it cannot be removed due to the fact that they are two different fields `inclusive_from` and `inclusive_to`.

Therefore, moving this `statement` higher in this case will not make sense.
Or did I miss something :) ?

vvasiloi: Right, my mistake.

oallain (approved):

NoResponseMate (approved):

NoResponseMate:
Thank you, @christopherhero!

vvasiloi:
Thank you @christopherhero, for having patience and finishing the work even after so much time passed! 🥇

PR #155, opened 2021-05-11
[WIP] Grid config builders

loic425:
This is linked to [Nyholm tweet](https://twitter.com/TobiasNyholm/status/1376106461391155200) and [Nyholm PR](https://github.com/symfony/symfony/pull/40600).

I think we have to create a pollyfill for the incoming interface from Symfony config.


src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,207 @@
...
+
+        return $output;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExtensionAlias(): string
+    {
+        return 'sylius_resource';
```
loic425, line 205: That must be removed.

src/Component/composer.json
```diff
@@ -31,6 +31,7 @@
         "php": "^7.4 || ^8.0",
         "sylius-labs/polyfill-symfony-event-dispatcher": "^1.0",
         "sylius/registry": "^1.5",
+        "symfony/config": "^4.4 || ^5.2",
```
lchrusciel, line 34: Perhaps all dependencies should be bumped to 5.3 according to github workflow, or we should increase amount of builds and test 4.4, 5.2 and 5.3

src/Component/spec/Config/Builder/FieldSpec.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
```
lchrusciel, line 2: Minor thing, but we usually add it 
```suggestion

declare(strict_types=1);

```

src/Component/spec/Config/Builder/FieldSpec.php
```diff
@@ -0,0 +1,82 @@
...
+namespace spec\Sylius\Component\Grid\Config\Builder;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Grid\Config\Builder\Field;
+
+class FieldSpec extends ObjectBehavior
```
lchrusciel, line 8:
```suggestion
final class FieldSpec extends ObjectBehavior
```
but that could be improved some day

src/Component/spec/Config/Builder/FieldSpec.php
```diff
@@ -0,0 +1,82 @@
...
+{
+    function let(): void
+    {
+        $this->beConstructedWith('name', 'string');
+    }
+
+    function it_is_initializable(): void
+    {
+        $this->shouldHaveType(Field::class);
```
lchrusciel, line 17: Can we also check the interface existance?
```suggestion
        $this->shouldImplement(FieldInterface::class);
```

src/Component/Config/Builder/Field.php
```diff
@@ -0,0 +1,120 @@
...
+    private ?bool $enabled = null;
+    /** @var bool|string|null */
+    private $sortable = null;
+    private ?int $position = null;
+    private array $options = [];
+
+    public function __construct(string $name, string $type)
+    {
+        $this->name = $name;
+        $this->type = $type;
```
lchrusciel, line 31: I'm just wondering if type should have predefined set of values

src/Bundle/test/config/packages/framework.yaml
```diff
@@ -10,3 +10,4 @@ framework:
         storage_id: session.storage.mock_file
     http_method_override: true
     test: ~
+    annotations: true
```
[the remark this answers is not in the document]
loic425: Yes, but not for this PR :)
It's already on master => https://github.com/Sylius/SyliusGridBundle/blob/master/src/Bundle/test/config/packages/framework.yaml

src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    public function addBulkAction(ActionInterface $action): self
+    {
+        $this->addAction($action, 'bulk');
+
+        return $this;
+    }
+
+    public function addCreateAction(array $options = [], string $group = 'main'): self
```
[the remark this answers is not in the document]
loic425: I think it's awesome! a very good idea to improve the way to add your custom predefined actions.
Tha's the power of PR reviews!

src/Component/composer.json
loic425: @lchrusciel yes and add a "^6.0" but in a separated PR, what do you think?

src/Bundle/test/config/sylius/grids/author.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+use App\Entity\Author;
+use Sylius\Component\Grid\Config\Builder\Field;
+use Sylius\Component\Grid\Config\Builder\Filter;
+use Sylius\Component\Grid\Config\Builder\GridBuilder;
+use Sylius\Component\Grid\Config\Builder\GridConfig;
+
+return static function (GridConfig $grid) {
+    $grid->addGrid(GridBuilder::create('app_author', Author::class)
```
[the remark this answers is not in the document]
loic425: What do you mean?

src/Component/Config/Builder/Field.php
loic425: Predefined set of values are on this next PR:
https://github.com/Sylius/SyliusGridBundle/pull/173

src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        $this->actionGroups[$group]->addAction($action);
+
+        return $this;
+    }
+
+    public function addMainAction(ActionInterface $action): self
+    {
+        $this->addAction($action, 'main');
```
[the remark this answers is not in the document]
loic425: Which one? GridBuilderInterface?

src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,225 @@
...
+
+        return $this;
+    }
+
+    public function addItemAction(ActionInterface $action): self
```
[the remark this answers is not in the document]
loic425: Done :)

src/Bundle/test/config/sylius/grids/author.php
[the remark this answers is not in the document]
lchrusciel: I'm wondering if we will be able to replace our current, YAML grid specifications with the new syntax. But as Sylius models will be changed in the end-user app, we need to be able to use parameters. E.g.
```suggestion
    $grid->addGrid(GridBuilder::create('app_author', '%app.model.author.class%')
```

src/Component/composer.json
lchrusciel: With Symfony 6.0 is a great idea to start testing. But in terms of this PR I see only two possibilities:
1) Bump all Symfony requirements to 5.3
2) Run pipelines for Symfony 4.4, 5.2, and 5.3
both of them are fine for me, but I would say that 1 is probably simpler so that would be my choice

src/Component/Config/Builder/GridBuilder.php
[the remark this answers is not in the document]
lchrusciel: I thought about `ActionGroupInterface`, but `GridBuilderInterface` makes sense as well.

```suggestion
        $this->addAction($action, ActionGroupInterface::MAIN_GROUP);
```

src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        return $this;
+    }
+
+    public function addAction(ActionInterface $action, string $group): self
+    {
+        if (!isset($this->actionGroups[$group])) {
+            $this->actionGroups[$group] = ActionGroup::create($group);
+        }
```
lchrusciel, line 108: Shouldn't we reuse addActionGroup
```suggestion
        $this-> addActionGroup($group)
```

src/Component/Config/Builder/GridBuilder.php
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    public function addFilter(FilterInterface $filter): self
+    {
+        $this->filters[$filter->getName()] = $filter;
+
+        return $this;
+    }
+
+    public function addActionGroup(string $name): self
```
lchrusciel, line 95: What would you say for the following contract?
```suggestion
    public function addActionGroup(ActionGroupInterface $name): self
```
This one would be more coherent with `addAction` function. What is more, it would allow us to have following syntax:
```php
$grid
    ->addActionGroup(ActionGroup::create('bulk'))
    ->addActionGroup(MainActionGroup::create(CreateAction::create()))
    ->addActionGroup(ItemActionGroup::create(UpdateAction::create(), DeleteAction::create())) // with variadics
;
```

src/Component/Config/Builder/Filter/Filter.php
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Config\Builder\Filter;
+
+class Filter implements FilterInterface
```
lchrusciel, line 16:
```suggestion
final class Filter implements FilterInterface
```

src/Bundle/test/config/sylius/grids/book_by_english_authors.php
```diff
@@ -0,0 +1,44 @@
...
+use App\Entity\Nationality;
+use Sylius\Component\Grid\Config\Builder\Field\Field;
+use Sylius\Component\Grid\Config\Builder\Filter\Filter;
+use Sylius\Component\Grid\Config\Builder\GridBuilder;
+use Sylius\Component\Grid\Config\Builder\GridConfig;
+
+return static function (GridConfig $grid) {
+    $grid->addGrid(GridBuilder::create('app_book_by_english_authors', Book::class)
+        ->setRepositoryMethod(["expr:service('app.english_books_query_builder')", 'create'])
+        ->addFilter(Filter::create('title', 'string'))
```
lchrusciel, line 13: All already existing filter types could be good candidates for consts. But let do this in separate PR (perhaps with predefined classes for them)

src/Component/Config/Builder/GridBuilder.php
loic425: @lchrusciel I added a variadic actions parameter on ActionGroup::create method cause it allows to have a simple MainActionGroup, ItemActionGroup, SubItemActionGroup and BulkActionGroup.

I created these four predefined action groups, but I think you prefer them in another PR 😄

src/Component/composer.json
loic425: @lchrusciel first choice done

src/Component/Config/Builder/GridBuilder.php
loic425: https://github.com/Sylius/SyliusGridBundle/pull/174

src/Component/Config/Builder/ActionGroup/ActionGroup.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    private string $name;
+    private array $actions = [];
+
+    private function __construct(string $name)
+    {
+        $this->name = $name;
+    }
+
+    public static function create(string $name, ...$actions): ActionGroupInterface
```
loic425, line 28:
```suggestion
    public static function create(string $name, array $actions = []): ActionGroupInterface
```

src/Component/Config/Builder/ActionGroup/ActionGroupInterface.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public const MAIN_GROUP = 'main';
+    public const ITEM_GROUP = 'item';
+    public const SUB_ITEM_GROUP = 'subitem';
+    public const BULK_GROUP = 'bulk';
+
+    /**
+     * @param ActionInterface[] $actions
+     */
+    public static function create(string $name, ...$actions): self;
```
loic425, line 28:
```suggestion
    public static function create(string $name, array $actions): self;
```

loic425:
@lchrusciel What do you think about adding these following methods?
* public function addMainActions(array $actions): self
* public function addItemActions(array $actions): self
* public function addSubItemActions(array $actions): self
* public function addBulkActions(array $actions): self

Maybe on (this PR)[#174]

.github/workflows/build.yml
```diff
@@ -74,3 +74,10 @@ jobs:
             -
                 name: Run bundle tests
                 run: composer test
+
+            -
+                name: Run tests for grids with php config
+                if: matrix.symfony == '^5.3'
+                run: |
+                    (cd src/Bundle/test && bin/console cache:clear --env=test_grids_with_php_config)
+                    APP_ENV=test_grids_with_php_config composer test
```
lchrusciel: I'm wondering if it is enough to test this new config style. The problem is that even tho, this may check the current implementation, we cannot be 100% sure. What I mean, is that you may remove the new config and the test will still pass (due to YAML configuration loaded in services.yaml). What we could do, is to have two envs `test_grids_with_php_config` and `test_grids_with_yaml_config`.

lchrusciel (approved):

src/Bundle/test/config/sylius/grids/book_by_english_authors.php
```diff
@@ -0,0 +1,44 @@
...
+            ->setLabel('Author')
+            ->setPath('author.name')
+            ->setSortable(true, 'author.name')
+        )
+        ->addField(Field::create('nationality', 'string')
+            ->setLabel('Nationality')
+            ->setPath('na.name')
+            ->setSortable(true, 'na.name')
+        )
+        ->setLimits([10, 5, 15])
```
Zales0123, line 42: As far as I see, we do not configure any actions in these test configurations. Maybe it would be worth giving them a try? Just to be sure :)

Zales0123 (approved):
This is spectacular 🎉 🚀

src/Bundle/test/config/sylius/grids/book_by_english_authors.php
loic425: @Zales0123 Do we think it could be enough to add actions without testing it via end-to-end tests?
I think it could validate the grid configuration.

mamazu:
I have a question about this. As far as I understand this this is only a different way to generate the Array that was "previously" in YAML. Then it should be possible to use the parameter `app.model.book` instead of the `Book::class` in all definitions right? Because otherwise this would make it hard for plugins to be easily extendible.

loic425:
> I have a question about this. As far as I understand this this is only a different way to generate the Array that was "previously" in YAML. Then it should be possible to use the parameter `app.model.book` instead of the `Book::class` in all definitions right? Because otherwise this would make it hard for plugins to be easily extendible.



> I have a question about this. As far as I understand this this is only a different way to generate the Array that was "previously" in YAML. Then it should be possible to use the parameter `app.model.book` instead of the `Book::class` in all definitions right? Because otherwise this would make it hard for plugins to be easily extendible.

Yes it's possible to use '%app.model.book.class%' instead. We added a test to check this syntax still works.

src/Bundle/Config/Builder/GridBuilderInterface.php
```diff
@@ -0,0 +1,53 @@
...
+
+    public function setLimits(array $limits): self;
+
+    public function addFilter(FilterInterface $filter): self;
+
+    public function removeFilter(string $name): self;
+
+    public function addActionGroup(ActionGroupInterface $actionGroup): self;
+
+    public function addAction(ActionInterface $action, string $group): self;
```
lchrusciel, line 50: Lets add remove methods as well, to be coherent with field and filter API
```suggestion
    public function addActionGroup(ActionGroupInterface $actionGroup): self;

    public function removeActionGroup(string $name): self;
    
    public function addAction(ActionInterface $action, string $group): self;
    
    public function removeAction(string $name, string $group): self;
```

src/Bundle/Tests/DependencyInjection/GridBuilderConfigurationTest.php
```diff
@@ -0,0 +1,353 @@
...
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_builds_grid_with_filters(): void
+    {
+        $gridBuilder = GridBuilder::create('app_admin_book', Book::class)
+            ->AddFilter(Filter::create('search', 'string'))
```
lchrusciel, line 96:
```suggestion
            ->addFilter(Filter::create('search', 'string'))
```

lchrusciel (approved):

PR #177, opened 2021-11-03
[GridBuilder] Docs

loic425:


docs/advanced_configuration.md
```diff
@@ -14,4 +16,27 @@ sylius_grid:
...
+
+```php
+<?php
+
+use Sylius\Bundle\GridBundle\Builder\Field\StringField;
+use Sylius\Bundle\GridBundle\Builder\GridBuilder;
+use Sylius\Bundle\GridBundle\Config\GridConfig;
+
+return static function (GridConfig $grid) {
+    $grid->addGrid(GridBuilder::create('app_user', '%app.model.user.class%')
```
lchrusciel, line 31: It would be perfect if these examples would configure the same stuff one-to-one. If it is not possible with grid, I would prefer to adjust YAML configuration to it

lchrusciel (approved):

lchrusciel:
Thank you, Loïc! :1st_place_medal:

PR #183, opened 2021-11-03
Tests with Postgres

loic425:


.github/workflows/build.yml
```diff
@@ -12,29 +12,41 @@ on:
...
 
                 include:
                     -
                         php: 7.4
                         symfony: ""
                         currency-bundle: "^1.7"
+                    -
+                        php: 8.0
+                        symfony: "^5.3"
+                        postgres: "13.3"
```
lchrusciel, line 31: I'm wondering if we shouldn't test v14 as well. Also, if we are testing with postgres, shouldn't we trigger build with MySQL? At the end, we are testing SqlLite only right now

lchrusciel (commented):
I'm a little bis sceptical about the current stage of this PR and if it is needed. Anyway, it requires rebase

lchrusciel:
Ok, I've was just convinced about need of this PR with #184

.github/workflows/build.yml
loic425: Yes I can add the v14 for Postgres, but I think the MySQL build can be added in another PR.

loic425: done for PostgreSQL v14

lchrusciel:
Thanks Loic! 🎉

PR #184, opened 2021-11-04
Fix like with PostgresSQL

loic425:


src/Bundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -129,14 +129,20 @@ public function like(string $field, string $pattern)
...
 
     public function notLike(string $field, string $pattern)
     {
         $field = $this->adjustField($field);
 
-        return $this->queryBuilder->expr()->notLike($this->resolveFieldByAddingJoins($field), $this->queryBuilder->expr()->literal($pattern));
+        return $this->queryBuilder->expr()->notLike(
+            (string) $this->queryBuilder->expr()->lower($this->resolveFieldByAddingJoins($field)),
+            $this->queryBuilder->expr()->literal(strtolower($pattern))
+        );
```
lchrusciel, line 145: We should have a test case for this case as well

lchrusciel (approved):

PR #185, opened 2021-11-04
[GridBuilder] Fix entity filter with multiple option

loic425:


lchrusciel (commented):
Can we also write sample e2e test for this case?

loic425:
> Can we also write sample e2e test for this case?

done

lchrusciel (approved):

PR #186, opened 2021-11-04
[GridBuilder] Add some missing features on field builder

loic425:


src/Bundle/spec/Builder/Field/TwigFieldSpec.php
```diff
@@ -22,4 +22,14 @@ function it_creates_fields(): void
         $field->shouldHaveType(FieldInterface::class);
         $field->getName()->shouldReturn('enabled');
     }
+
+    function it_creates_fields_with_vars(): void
```
loic425, line 26: @lchrusciel This is the reason for this PR, when using TwigField, you might want to defines some other options.

loic425: @lchrusciel I'm wondering if an array_merge instead of replacing all options (and loose some predefined options) could be better for end users.

lchrusciel: `addOption` behaves more like `setOption`. What you described is something like `addOptions`

loic425: now we have
* setOption
* setOptions
* addOptions
* removeOption

lchrusciel (approved):

PR #188, opened 2021-11-05
Fixed template syntax in route config

veloxy:
Spotted old template syntax in one of the docs which no longer works


lchrusciel (approved):
Thanks for your contribution Kevin :)

PR #190, opened 2021-11-08
[GridBuilder] Allow using class as parameter for grids as service

loic425:
With a resource class as parameter, it already works like this code below

```yaml
services:
    _defaults:
        bind:
            $productClass: '%sylius.model.product.class%'
```

and using driver option

```php
<?php

declare(strict_types=1);

namespace App\Grid;

use App\Entity\Product\Product;
use Sylius\Bundle\GridBundle\Grid\AbstractGrid;

final class AdminProductGrid extends AbstractGrid
{
    private string $productClass;

    public function __construct(string $productClass)
    {
        $this->productClass = $productClass;
    }

    public static function getName(): string
    {
        return 'sylius_admin_product';
    }

    public function buildGrid(GridBuilderInterface $gridBuilder): void
    {
        $gridBuilder
            ->setDriverOption('class', $this->productClass)
        ;
    }
}
```

But I think it's better like this

```php
<?php

declare(strict_types=1);

namespace App\Grid;

use App\Entity\Product\Product;
use Sylius\Bundle\GridBundle\Grid\AbstractGrid;

final class AdminProductGrid extends AbstractGrid
{
    private string $productClass;

    public function __construct(string $productClass)
    {
        $this->productClass = $productClass;
    }

    public static function getName(): string
    {
        return 'sylius_admin_product';
    }

    public function getResourceClass(): string
    {
        return $this->productClass;
    }

    public function buildGrid(GridBuilderInterface $gridBuilder): void
    {
    }
}
```

But it requires to remove the "static" for the getResourceClass method.

### DIFF
--- a/src/Bundle/Grid/AbstractGrid.php
+++ b/src/Bundle/Grid/AbstractGrid.php
@@ -30,7 +30,7 @@ abstract class AbstractGrid implements GridInterface
     private function createGridBuilder(): GridBuilderInterface
     {
         if ($this instanceof ResourceAwareGridInterface) {
-            return GridBuilder::create($this::getName(), $this::getResourceClass());
+            return GridBuilder::create($this::getName(), $this->getResourceClass());
         }
 
         return GridBuilder::create($this::getName());
--- a/src/Bundle/test/config/services_test_grids_as_service.yaml
+++ b/src/Bundle/test/config/services_test_grids_as_service.yaml
@@ -2,6 +2,8 @@ services:
     _defaults:
         autowire: true
         autoconfigure: true
+        bind:
+            $authorClass: '%app.model.author.class%'
 
     App\QueryBuilder\EnglishBooksQueryBuilder: null
 
--- a/src/Component/Tests/Dummy/ClassAsParameterGrid.php
+++ b/src/Component/Tests/Dummy/ClassAsParameterGrid.php
@@ -1,14 +1,5 @@
 <?php
 
-/*
- * This file is part of the Sylius package.
- *
- * (c) Paweł Jędrzejewski
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
 declare(strict_types=1);
 
 namespace Sylius\Component\Grid\Tests\Dummy;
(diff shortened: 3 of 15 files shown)


src/Bundle/Grid/AbstractGrid.php
```diff
@@ -30,7 +30,7 @@ public function toArray(): array
     private function createGridBuilder(): GridBuilderInterface
     {
         if ($this instanceof ResourceAwareGridInterface) {
-            return GridBuilder::create($this::getName(), $this::getResourceClass());
+            return GridBuilder::create($this::getName(), $this->getResourceClass());
```
lchrusciel, line 33: if I understand correctly, this class(parent of `AbstractGrid`) is already initiated, am I correct?

src/Bundle/test/config/services_test_grids_as_service.yaml
```diff
@@ -2,6 +2,8 @@ services:
     _defaults:
         autowire: true
         autoconfigure: true
+        bind:
+            $authorClass: '%app.model.author.class%'
```
lchrusciel, line 6: Perhaps it could be improved with ResourceBundle to create such binds out-of-the-box

src/Component/Tests/Dummy/ClassAsParameterGrid.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public static function getName(): string
+    {
+        return 'app_class_as_parameter';
+    }
+
+    public function getResourceClass(): string
+    {
+        return $this->authorClass;
```
lchrusciel, line 27: only better it could work, if we would be able to use parameter directly here. But really neat improvement

lchrusciel (approved):

src/Bundle/test/config/services_test_grids_as_service.yaml
loic425: Yes I had this same idea during development.

src/Component/Tests/Dummy/ClassAsParameterGrid.php
loic425: Yes, but I don't know how...

PR #191, opened 2021-11-09
[GridBuilder] Fix extending a grids as service

loic425:


src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -1,305 +0,0 @@
...
-declare(strict_types=1);
-
-namespace Sylius\Bundle\GridBundle\Tests\DependencyInjection;
-
-use Matthias\SymfonyConfigTest\PhpUnit\ConfigurationTestCaseTrait;
-use PHPUnit\Framework\TestCase;
-use Sylius\Bundle\GridBundle\DependencyInjection\Configuration;
-use Sylius\Bundle\GridBundle\Doctrine\ORM\Driver;
-
-final class ConfigurationTest extends TestCase
```
lchrusciel, line 21: Why did you removed this class?

src/Component/Tests/Dummy/FooFightersGrid.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Tests\Dummy;
+
+use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
+use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
+use Sylius\Bundle\GridBundle\Grid\ResourceAwareGridInterface;
+
+final class FooFightersGrid extends AbstractGrid implements ResourceAwareGridInterface
```
lchrusciel, line 20: Can we write one e2e test with usage of this grid?

src/Bundle/Provider/ServiceGridProvider.php
```diff
@@ -40,6 +41,27 @@ public function get(string $code): Grid
...
+
+    private function extend(array $gridConfiguration, string $parentGridCode): array
+    {
+        $parentGrid = $this->gridRegistry->getGrid($parentGridCode);
+
+        Assert::notNull($parentGrid, sprintf('Parent grid with code "%s" does not exists.', $parentGridCode));
+
+        $parentGridConfiguration = $parentGrid->toArray();
+
+        unset($parentGridConfiguration['sorting']); // Do not inherit sorting.
```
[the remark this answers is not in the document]
loic425: AHAH, cause it's on ArrayGridProvider.

src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
loic425: Oups, It should be an error.

src/Component/Tests/Dummy/FooFightersGrid.php
loic425: https://github.com/Sylius/SyliusGridBundle/pull/191/files#diff-19a4971649a951cf9ff2533feef055eca83186b5e7837011f6ff746717ca5ec9R694

loic425: But this one is better
https://github.com/Sylius/SyliusGridBundle/pull/191/files#diff-19a4971649a951cf9ff2533feef055eca83186b5e7837011f6ff746717ca5ec9R692

loic425: But this one is better https://github.com/Sylius/SyliusGridBundle/pull/191/files#diff-e5f9e1c0731852fd99ea54b11133276943457f440a8ff43f4ee9b4a4a1831915R23

src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
loic425: fixed

src/Bundle/Provider/ServiceGridProvider.php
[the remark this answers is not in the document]
lchrusciel: I don't get it, so what is the issue with that? Do we have some case, when it fails?

lchrusciel (approved):

src/Component/Tests/Dummy/FooFightersGrid.php
lchrusciel: But we don't change anything there. Shouldn't we try to turn off some field and add another?

loic425: @lchrusciel done https://github.com/Sylius/SyliusGridBundle/pull/191/commits/1e48c9a8b3d5fb57312bda7f66f0568fc55b7106

src/Bundle/spec/Provider/ServiceGridProviderSpec.php
```diff
@@ -43,11 +43,44 @@ function it_gets_grids_definitions_by_code(
...
 
+    function it_supports_grid_inheritance(
+        ArrayToDefinitionConverterInterface $converter,
+        GridRegistryInterface $gridRegistry,
+        GridInterface $fooGrid,
+        GridInterface $fooFightersGrid,
+        Grid $fooGridDefinition,
+        Grid $fooFightersGridDefinition
+    ): void
+    {
```
lchrusciel, line 54:
```suggestion
    ): void {
```

lchrusciel (approved):

Zales0123 (approved):

src/Bundle/Provider/ServiceGridProvider.php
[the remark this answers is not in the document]
loic425: I used a Trait, I think this is easy like this without using inheritance for that or a tiny service just for that.

src/Bundle/Provider/ServiceGridProvider.php
```diff
@@ -40,6 +41,27 @@ public function get(string $code): Grid
...
+        }
+
         return $this->converter->convert($code, $gridConfiguration);
     }
+
+    private function extend(array $gridConfiguration, string $parentGridCode): array
+    {
+        $parentGrid = $this->gridRegistry->getGrid($parentGridCode);
+
+        Assert::notNull($parentGrid, sprintf('Parent grid with code "%s" does not exists.', $parentGridCode));
```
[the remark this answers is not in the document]
loic425: done

src/Bundle/Provider/ServiceGridProvider.php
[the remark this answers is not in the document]
loic425: refatored with a tiny service.

src/Component/Provider/ArrayGridProvider.php
```diff
@@ -37,22 +44,20 @@ public function get(string $code): Grid
...
 
         $gridConfiguration = $this->gridConfigurations[$code];
+        $parentGridCode = $gridConfiguration['extends'] ?? null;
+
+        if (null !== $parentGridCode) {
+            $parentGridConfiguration = $this->gridConfigurations[$gridConfiguration['extends']] ?? null;
 
-        if (isset($gridConfiguration['extends'], $this->gridConfigurations[$gridConfiguration['extends']])) {
-            $gridConfiguration = $this->extend($gridConfiguration, $this->gridConfigurations[$gridConfiguration['extends']]);
+            Assert::notNull($parentGridConfiguration, sprintf('Parent grid with code "%s" does not exists.', $gridConfiguration['extends']));
```
vvasiloi, line 52: This might break someone's code... it needs a mention in the UPGRADE file.

src/Component/Provider/ArrayGridProvider.php
```diff
@@ -37,22 +44,20 @@ public function get(string $code): Grid
...
     private function extend(array $gridConfiguration, array $parentGridConfiguration): array
     {
-        unset($parentGridConfiguration['sorting']); // Do not inherit sorting.
-
-        $configuration = array_replace_recursive($parentGridConfiguration, $gridConfiguration) ?: [];
-
-        unset($configuration['extends']);
-
-        return $configuration;
+        return $this->gridConfigurationExtender->extends($gridConfiguration, $parentGridConfiguration);
```
vvasiloi, line 61: This method can be removed, it doesn't do anything.

src/Component/spec/Configuration/GridConfigurationExtenderSpec.php
```diff
@@ -0,0 +1,44 @@
...
+    function it_does_not_extend_sorting_configuration(): void
+    {
+        $gridConfiguration = ['foo' => 'fighters'];
+        $parentGridConfiguration = ['sorting' => ['name' => 'asc']];
+
+        $this->extends($gridConfiguration, $parentGridConfiguration)->shouldReturn([
+            'foo' => 'fighters',
+        ]);
+    }
+}
```
vvasiloi, line 44: Maybe also test that it removed the `exdends` key.

src/Component/Provider/ArrayGridProvider.php
loic425: done

src/Component/Provider/ArrayGridProvider.php
loic425: done

src/Component/spec/Configuration/GridConfigurationExtenderSpec.php
loic425: done

vvasiloi (approved):

src/Component/Provider/ArrayGridProvider.php
```diff
@@ -13,21 +13,28 @@
...
 
     /** @var array[] */
     private array $gridConfigurations;
 
-    public function __construct(ArrayToDefinitionConverterInterface $converter, array $gridConfigurations)
-    {
+    public function __construct(
+        ArrayToDefinitionConverterInterface $converter,
+        array $gridConfigurations,
+        GridConfigurationExtenderInterface $gridConfigurationExtender
```
lchrusciel, line 33: This argument should be nullable to maintain BC

src/Bundle/Provider/ServiceGridProvider.php
```diff
@@ -40,6 +48,21 @@ public function get(string $code): Grid
...
+
+        $parentGridConfiguration = $parentGrid->toArray();
+
+        return $this->gridConfigurationExtender->extends($gridConfiguration, $parentGridConfiguration);
+    }
```
lchrusciel, line 67: Can't we extract the whole part to separate service? Therefore, we would throw an exception only in one place - and for whatever reason it would be easier for others to replace/change this logic

vvasiloi: You mean to only pass the parent grid code to the extender and use the grid registry to get the grid by code and throw exception if it doesn't exists?

lchrusciel: Something like this. The content of this method seems to be almost duplicated and I assume, we would like to use the same logic if we will need to use this function again

vvasiloi: It's similar, but not the same.
Here, the grid registry is used, while the ArrayGridProvider uses an array copy of grid configuration, that's where it gets it's name from.
@loic425 are those two compatible? Can a grid registered as service extend a grid from config? What about the other way around?

src/Component/Provider/ArrayGridProvider.php
loic425: done

src/Bundle/Provider/ServiceGridProvider.php
loic425: In theory, with "extends" method you can extend a grid from config with a grid registered as service but not the other way around.

The solution on the service provider is to look on the grid registry to find the grid configuration. The grid registry has its providers sorted by priority (by default, ServiceGridProvider is on first position).

It's impossible for now, cause GridRegistry is on the bundle part and the ArrayGridProvider is on Component part.

loic425: At first, I added all these services into Component :D and made the DI into the bundle part.

vvasiloi: Let's address this separately.

loic425: @lchrusciel Ok for you too?

lchrusciel: Ok, I'm fine with that 👍

lchrusciel:
Thanks Loic! 🎉

PR #143, opened 2021-03-19
Pagerfanta updates

mbabker:
Very simple PR which:

- Updates the used adapter classes to stop using the deprecated classes
- Allows Pagerfanta 3.0

PR #204, opened 2022-01-10
Add a conflict for pagerfanta 3.0 before allowing it in #143

loic425:
On Sylius\Bundle\GridBundle\Docctrine\DBAL\DataSource, there is a usage of Pagerfanta\Adapter\DoctrineDbalAdapter which does not exist on pagerfanta 3.0.

It will be replaced on #143 
https://github.com/Sylius/SyliusGridBundle/blob/5da00cdd4916473025de064822332ed5144c959f/src/Bundle/Doctrine/DBAL/DataSource.php#L17

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,10 @@
         "sylius/grid": "self.version"
     },
     "conflict": {
-        "pagerfanta/pagerfanta": "^3.0",
+        "pagerfanta/doctrine-dbal-adapter": ">=3.0",
+        "pagerfanta/doctrine-orm-adapter": ">=3.0",
+        "pagerfanta/doctrine-phpcr-odm-adapter": ">=3.0",
+        "pagerfanta/pagerfanta": ">=3.0",
         "twig/twig": "^1.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,7 @@
         "sylius/grid": "self.version"
     },
     "conflict": {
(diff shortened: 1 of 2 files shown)


vvasiloi (approved):

lchrusciel (commented):
Lets go with this solution, but we should introduce CONFLICT.md file with some explanation

composer.json
```diff
@@ -46,6 +46,7 @@
         "sylius/grid": "self.version"
     },
     "conflict": {
+        "pagerfanta/pagerfanta": "^3.0",
```
mbabker, line 49: You need to conflict the individual packages as well, after the Pagerfanta 2.4 releases there is no guarantee that downstream users are installing the `pagerfanta/pagerfanta` mono-package.

Based on the issue report, `pagerfanta/doctrine-dbal-adapter` needs to be added at a minimum.

You'll also need a conflict blocking versions older than 2.4 as that's when the newer classes were introduced.

loic425: @mbabker Oh you're right.

loic425: @mbabker Do you want to make another PR replacing this one?

mbabker (approved):

loic425:
@lchrusciel I added the CONFLICTS.md to explain the added conflicts.

GSadee (approved):

lchrusciel:
Closing in favour of #205

PR #206, opened 2022-01-26
Make grid

loic425:
Replaces #46


lbaey:
👍

src/Bundle/Maker/MakeGrid.php
```diff
@@ -0,0 +1,167 @@
...
+        );
+
+        $generator->writeChanges();
+
+        $this->writeSuccessMessage($io);
+    }
+
+    public function configureDependencies(DependencyBuilder $dependencies): void
+    {
+        // noop
```
lchrusciel, line 121: ?

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        $tester = new CommandTester((new Application(self::bootKernel()))->find('make:grid'));
+
+        $this->assertFileDoesNotExist(self::tempFile('Grid/PriceGrid.php'));
+
+        $tester->execute(['entity' => Price::class, '--namespace' => 'Tests\Tmp\Grid']);
+
+        $this->assertFileExists(self::tempFile('Grid/PriceGrid.php'));
+        $this->assertSame(self::getPriceGridExpectedContent(), \file_get_contents(self::tempFile('Grid/PriceGrid.php'))
+        );
```
lchrusciel, line 34:
```suggestion
        $this->assertSame(self::getPriceGridExpectedContent(), \file_get_contents(self::tempFile('Grid/PriceGrid.php')));
```

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -0,0 +1,143 @@
...
+    /** @test */
+    public function it_can_create_grids(): void
+    {
+        $tester = new CommandTester((new Application(self::bootKernel()))->find('make:grid'));
+
+        $this->assertFileDoesNotExist(self::tempFile('Grid/PriceGrid.php'));
+
+        $tester->execute(['entity' => Price::class, '--namespace' => 'Tests\Tmp\Grid']);
+
+        $this->assertFileExists(self::tempFile('Grid/PriceGrid.php'));
```
lchrusciel, line 32: Can we extract `'Grid/PriceGrid.php'` to const, so we will be sure that the path is identical?

src/Bundle/Resources/config/skeleton/Grid.tpl.php
```diff
@@ -0,0 +1,85 @@
...
+        echo "                    ->setLabel('".ucfirst($fieldname)."')\n";
+        echo "             )\n";
+    }
+
+    if ('BOOLEAN' === $type) {
+        echo "            ->addField(\n";
+        echo "                TwigField::create('".$fieldname."', 'path/to/field/template.html.twig')\n";
+        echo "                    ->setLabel('".ucfirst($fieldname)."')\n";
+        echo "             )\n";
+    }
```
lchrusciel, line 58: I'm not sure about this node. What is so special about boolean node, that it is handled with TwigField? I know, that this is how we do it mostly in Sylius, but the current behaviour will result in 500 error, as the template path does not exist. I would use a string field to be sure, that the generated grid can work just after the generation

.gitignore
```diff
@@ -4,3 +4,4 @@
 /composer.lock
 
 /.phpunit.result.cache
+src/Bundle/test/tmp
```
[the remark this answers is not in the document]
loic425: The directory is automatically created by Symfony maker command (I guess using symfony/filesystem), so it is not necessary.

src/Bundle/Maker/MakeGrid.php
loic425: It's required by the interface `Symfony\Bundle\MakerBundle\MakerInterface'.
But I have to check the return typehint on Symfony 6.

src/Bundle/Resources/config/skeleton/Grid.tpl.php
loic425: Yes you're right, we need a generated grid which work just after the genration. I agree.

Maybe I should comment this code lines just to show you can handle it with a twig template.
and maybe add a comment with the "@SyliusUi/Grid/Field/enabled.html" for Sylius/Monofony users too...

loic425: WDYT?

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
loic425: Good idea, on this class or in MarkerTestCase?

src/Bundle/Maker/MakeGrid.php
loic425: No that's good, that's not part of Symfony monorepository
https://github.com/symfony/maker-bundle/blob/main/src/MakerInterface.php

loic425: https://github.com/symfony/maker-bundle/blob/006577909ded549ee48fd69aa4da1a75ff9a2b55/src/Maker/MakeSerializerNormalizer.php

src/Bundle/Resources/config/skeleton/Grid.tpl.php
mbabker: Better question.  Why doesn't this bundle provide a `BooleanField` that's usable out-of-the-box?  A boolean field is a pretty common use case, it shouldn't need a Twig field to handle rendering unless you're aiming to do some custom display logic that a "simple" configuration couldn't handle ("simple" as in options to control the output text for both states, if you want to start using badges or icons for output that's where a Twig field is better suited).

loic425: That's a good reflexion, not for this PR, but for the future of this bundle.

loic425: But for sure, with TwigField (if we use it for other Doctrine types) have to be code-commented to keep the grid working just after the generation.

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
loic425: Done on this class, cause it's grid related. If we add some other makers on this bundle, it shouldn't be related.

src/Bundle/Maker/MakeGrid.php
lchrusciel: I meant - what do you mean by noop? No operation? I have the feeling, that both of us should be more specific :) I with my question and you with the comment here, so people would not need to try to resolve what is `noop`

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -0,0 +1,258 @@
...
+
+use App\Entity\Book;
+use App\Entity\Price;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Bundle\MakerBundle\Exception\RuntimeCommandException;
+use Symfony\Component\Console\Tester\CommandTester;
+
+final class MakeGridTest extends MakerTestCase
+{
+    public const BOOK_GRID_PATH = 'Grid/BookGrid.php';
```
lchrusciel, line 24: Do we need them to be public? 
```suggestion
    private const BOOK_GRID_PATH = 'Grid/BookGrid.php';
```

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -57,6 +59,10 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->registerForAutoconfiguration(GridInterface::class)
             ->addTag('sylius.grid')
         ;
+
+        if (!\class_exists(AbstractMaker::class)) {
```
lchrusciel, line 63: Can we somehow write test case for it? I'm not sure. The best idea I have right now is to duplicate test app without maker bundle... Did you test it manually?

src/Bundle/Maker/MakeGrid.php
loic425: I don't know lol
That's just copied/paste from https://github.com/zenstruck/foundry/blob/master/src/Bundle/Maker/MakeFactory.php#L170

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
loic425: I made that change on the three constants.

src/Bundle/DependencyInjection/SyliusGridExtension.php
loic425: I didn't test it manually, but I have an idea....

loic425: Voilà!
https://github.com/Sylius/SyliusGridBundle/pull/206/commits/b76ec9fcfab96d32fde3d752b08851231124e3b8

loic425: done... it wasn't as simple as I thought 🙄

lchrusciel (approved):

src/Bundle/Resources/config/skeleton/Grid.tpl.php
```diff
@@ -0,0 +1,85 @@
...
+}
+?>
+            ->addActionGroup(
+                MainActionGroup::create(
+                    CreateAction::create(),
+                )
+            )
+            ->addActionGroup(
+                ItemActionGroup::create(
+                    // ShowAction::create(),
```
lchrusciel, line 68: I would at least leave the comment why do we have it commented out

src/Bundle/Resources/config/skeleton/Grid.tpl.php
```diff
@@ -0,0 +1,85 @@
...
+
+    if (str_starts_with($type, 'DATE')) {
+        echo "            ->addField(\n";
+        echo "                DateTimeField::create('".$fieldname."')\n";
+        echo "                    ->setLabel('".ucfirst($fieldname)."')\n";
+        echo "             )\n";
+    }
+
+    if ('BOOLEAN' === $type) {
+        echo "             //->addField(\n";
```
lchrusciel, line 54: I would at least leave the comment why do we have it commented out

lchrusciel (approved):

lchrusciel:
Thank you, Loïc! :tada:

PR #209, opened 2022-02-04
Hotfix with datasources

loic425:
This is a hotfix for the "v1.11-ALPHA.1". The paginator does not work anymore due to a new "int" typehint for the current page on Pagerfanta 3.

### DIFF
--- a/src/Bundle/Doctrine/DBAL/DataSource.php
+++ b/src/Bundle/Doctrine/DBAL/DataSource.php
@@ -53,6 +53,8 @@ final class DataSource implements DataSourceInterface
 
     public function getData(Parameters $parameters)
     {
+        $page = (int) $parameters->get('page', 1);
+
         $countQueryBuilderModifier = function (QueryBuilder $queryBuilder): void {
             $queryBuilder
                 ->select('COUNT(DISTINCT o.id) AS total_results')
@@ -62,7 +64,7 @@ final class DataSource implements DataSourceInterface
 
         $paginator = new Pagerfanta(new QueryAdapter($this->queryBuilder, $countQueryBuilderModifier));
         $paginator->setNormalizeOutOfRangePages(true);
-        $paginator->setCurrentPage((int) $parameters->get('page', 1));
+        $paginator->setCurrentPage($page > 0 ? $page : 1);
 
         return $paginator;
     }
--- a/src/Bundle/Doctrine/DBAL/DataSource.php
+++ b/src/Bundle/Doctrine/DBAL/DataSource.php
@@ -62,7 +62,7 @@ final class DataSource implements DataSourceInterface
 
         $paginator = new Pagerfanta(new QueryAdapter($this->queryBuilder, $countQueryBuilderModifier));
         $paginator->setNormalizeOutOfRangePages(true);
-        $paginator->setCurrentPage($parameters->get('page', 1));
+        $paginator->setCurrentPage((int) $parameters->get('page', 1));
 
         return $paginator;
     }
--- a/src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
+++ b/src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
@@ -22,17 +22,19 @@ use Sylius\Component\Grid\Parameters;
 
 final class DataSourceSpec extends ObjectBehavior
 {
-    function it_implements_data_source(QueryBuilder $queryBuilder): void
+    function let(QueryBuilder $queryBuilder): void
     {
         $this->beConstructedWith($queryBuilder);
+    }
 
+    function it_implements_data_source(): void
+    {
         $this->shouldImplement(DataSourceInterface::class);
     }
 
     function it_gets_the_data(QueryBuilder $queryBuilder): void
     {
         $queryBuilder->getType()->willReturn(QueryBuilder::SELECT);
-        $this->beConstructedWith($queryBuilder);
 
         $this->getData(new Parameters(['page' => '1']))->shouldHaveType(Pagerfanta::class);
     }
--- a/src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
+++ b/src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\GridBundle\Doctrine\DBAL;
+
+use Doctrine\DBAL\Query;
+use Doctrine\DBAL\Query\QueryBuilder;
+use Pagerfanta\Pagerfanta;
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Parameters;
+
+final class DataSourceSpec extends ObjectBehavior
+{
+    function it_implements_data_source(
+        QueryBuilder $queryBuilder
+    ): void
+    {
+        $this->beConstructedWith($queryBuilder);
+
+        $this->shouldImplement(DataSourceInterface::class);
+    }
+
+    function it_should_get_the_data(QueryBuilder $queryBuilder): void
+    {
+        $queryBuilder->getType()->willReturn(QueryBuilder::SELECT);
+        $this->beConstructedWith($queryBuilder);
+
+        $this->getData(new Parameters(['page' => '1']))->shouldHaveType(Pagerfanta::class);
+    }
+}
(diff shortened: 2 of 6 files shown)


src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
```diff
@@ -0,0 +1,41 @@
...
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Parameters;
+
+final class DataSourceSpec extends ObjectBehavior
+{
+    function it_implements_data_source(
+        QueryBuilder $queryBuilder
+    ): void
+    {
```
GSadee, line 28:
```suggestion
    function it_implements_data_source(QueryBuilder $queryBuilder): void
    {
```

src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->beConstructedWith($queryBuilder);
+
+        $this->shouldImplement(DataSourceInterface::class);
+    }
+
+    function it_should_get_the_data(QueryBuilder $queryBuilder): void
```
GSadee, line 34:
```suggestion
    function it_gets_the_data(QueryBuilder $queryBuilder): void
```

src/Bundle/spec/Doctrine/ORM/DataSourceSpec.php
```diff
@@ -0,0 +1,39 @@
...
+use Pagerfanta\Pagerfanta;
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Parameters;
+
+final class DataSourceSpec extends ObjectBehavior
+{
+    function it_implements_data_source(
+        QueryBuilder $queryBuilder
+    ): void
```
GSadee, line 26:
```suggestion
    function it_implements_data_source(QueryBuilder $queryBuilder): void
```

src/Bundle/spec/Doctrine/ORM/DataSourceSpec.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->beConstructedWith($queryBuilder, false, false);
+
+        $this->shouldImplement(DataSourceInterface::class);
+    }
+
+    function it_should_get_the_data(QueryBuilder $queryBuilder): void
```
GSadee, line 33:
```suggestion
    function it_gets_the_data(QueryBuilder $queryBuilder): void
```

src/Bundle/spec/Doctrine/ORM/DataSourceSpec.php
```diff
@@ -0,0 +1,39 @@
...
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Parameters;
+
+final class DataSourceSpec extends ObjectBehavior
+{
+    function it_implements_data_source(
+        QueryBuilder $queryBuilder
+    ): void
+    {
+        $this->beConstructedWith($queryBuilder, false, false);
```
GSadee, line 28: Could be extracted to `let` method

GSadee (approved):

src/Bundle/spec/Doctrine/ORM/DataSourceSpec.php
loic425: Yes you're right.

loic425: done

src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->beConstructedWith($queryBuilder);
+
+        $this->shouldImplement(DataSourceInterface::class);
+    }
+
+    function it_gets_the_data(QueryBuilder $queryBuilder): void
+    {
+        $queryBuilder->getType()->willReturn(QueryBuilder::SELECT);
+        $this->beConstructedWith($queryBuilder);
```
Zales0123, line 35: `$this->beConstructedWith($queryBuilder);` can also be extracted to `let` method 🖖

src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
```diff
@@ -0,0 +1,39 @@
...
+
+        $this->getData(new Parameters(['page' => '1']))->shouldHaveType(Pagerfanta::class);
```
Zales0123, line 37: To be honest, it does not check too much :( Probably the best option would be to create a [custom matcher](https://github.com/Sylius/SyliusResourceBundle/blob/0bb5ab5dd36edc2e6402d22f0431f2d600efe488/src/Component/spec/Generator/RandomnessGeneratorSpec.php#L52) that would check if the returned object:

- is type of `Pagerfanta`
- has `$normalizeOutOfRangePages` set properly
- has `$currentPage` set as it should be

Then we're quite safe from regression :)

Zales0123 (commented):
Can you also explain why this change is needed in the PR description? Just for the record, if one would be looking for the reason of such hotfix, it needs to be explained somewhere :) 🖖

src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
loic425: done, I did not use the custom matcher, but I tried to and was not be satisfied.

src/Bundle/spec/Doctrine/DBAL/DataSourceSpec.php
loic425: done

src/Bundle/spec/Doctrine/PHPCRODM/DataSourceSpec.php
```diff
@@ -103,7 +103,11 @@ function it_should_get_the_data(
         $query->setFirstResult(Argument::any())->willReturn($query);
         $query->execute()->willReturn([]);
 
-        $this->getData(new Parameters(['page' => 1]))->shouldHaveType(Pagerfanta::class);
+        $data = $this->getData(new Parameters(['page' => '1']));
+
+        $data->shouldHaveType(Pagerfanta::class);
+        $data->getCurrentPage()->shouldReturn(1);
+        $data->getNormalizeOutOfRangePages()->shouldReturn(false);
```
loic425, line 110: @Zales0123 It has "false" value, is that normal?

src/Bundle/Doctrine/DBAL/DataSource.php
```diff
@@ -62,7 +64,7 @@ public function getData(Parameters $parameters)
 
         $paginator = new Pagerfanta(new QueryAdapter($this->queryBuilder, $countQueryBuilderModifier));
         $paginator->setNormalizeOutOfRangePages(true);
-        $paginator->setCurrentPage($parameters->get('page', 1));
+        $paginator->setCurrentPage($page > 0 ? $page : 1);
```
lchrusciel, line 67: Did we wanted to cover some specific case with this comparison? We didn't cover it with tests

src/Bundle/spec/Doctrine/PHPCRODM/DataSourceSpec.php
lchrusciel: This is current state of this data source, so I would leave it as is. Probably no one is using it or the current behaviour is ok, as there weren't any issues about it. We could also skip this check in this case

loic425: I think we can leave this test to be aware of the current implementation.

src/Bundle/Doctrine/DBAL/DataSource.php
loic425: Yes I can add a test with that behaviour. This has been added to fix a Psalm error.

lchrusciel: If so, I'm fine with it, but still would be better to cover with additional test I suppose

lchrusciel (approved):

lchrusciel:
Thanks, Loïc! :1st_place_medal:

src/Bundle/Doctrine/DBAL/DataSource.php
mbabker: Just for reference, the comparison here does two things:

1) Prevents Pagerfanta's `LessThan1CurrentPageException` from bubbling up if someone passes a negative number or 0 through the page parameter
2) Pagerfanta 3.x is a lot more strictly typed, both in terms of PHP and static analysis tools (I haven't figured out Psalm yet, but PHPStan is used pretty heavily on my repos, and both of those tools parse each other's doc block annotations which is a good thing).  This specific method has a `@phpstan-param positive-int` for the page parameter, so SA is doing the right thing by flagging it if it can't guarantee it's a positive integer, and the exception at runtime for invalid input is still there for less strictly typed code.

PR #214, opened 2022-02-14
[Docs] Resorting pages on docs homepage

loic425:
It seems like the pages order is weird on documentations. So I propose this new one.


Roshyo (approved):
I agree with this. Feels more natural for a regular learning path

Zales0123 (approved):

Zales0123:
Thank you, Loïc! :tada:

PR #217, opened 2022-02-15
[Docs] Remove invalid configuration for grid actions

Roshyo:
If you try to add icons to those actions, it will simply throw an exception since it does not expect null but string:

image

But even if you define a Semantic icon, it ignores it since the template does not expect it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -81,28 +81,24 @@ sylius_grid:
                         type: create
                         label: sylius.ui.create
                         enabled: true
-                        icon: ~
                         position: 100
                 item:
                     update:
                         type: update
                         label: sylius.ui.edit
                         enabled: true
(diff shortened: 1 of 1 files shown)


lchrusciel:
Hey Stephane! 

thanks for PR. However, according to our config: https://github.com/Sylius/SyliusGridBundle/blob/master/src/Bundle/DependencyInjection/Configuration.php#L152 this line is correct, therefore it should stay in the configuration reference. The problem is with this definition: https://github.com/Sylius/SyliusGridBundle/blob/master/src/Component/Definition/Action.php#L83 I suppose. Nonetheless, I'm not sure if Sylius will even use it. But if it doesn't that is something more to add as a warning. 

If you wish to fix it, probably some test for that would be nice :)

PR #219, opened 2022-02-15
[DX] [Docs] Adding more typing to FieldType and in doc

Roshyo:
I don't think it is a BC break introduction since the DocBlock was already announcing the type. But I'm not sure

### DIFF
--- a/src/Component/FieldTypes/FieldTypeInterface.php
+++ b/src/Component/FieldTypes/FieldTypeInterface.php
@@ -26,7 +26,7 @@ interface FieldTypeInterface
      *
      * @return string
      */
-    public function render(Field $field, $data, array $options);
+    public function render(Field $field, $data, array $options): string;
 
     /**
      * Configure options for this field type.
--- a/src/Component/FieldTypes/FieldTypeInterface.php
+++ b/src/Component/FieldTypes/FieldTypeInterface.php
@@ -23,8 +23,10 @@ interface FieldTypeInterface
      * $options.
      *
      * @param mixed $data
+     *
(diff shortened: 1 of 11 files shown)


loic425 (dismissed):
I agree with the documentation part, but for the render typehint, I'm pretty sure it is bc-break.

Roshyo:
In the interface, probably indeed :thinking: But we can still add that to the final classes imo.

src/Component/FieldTypes/FieldTypeInterface.php
```diff
@@ -23,10 +23,8 @@ interface FieldTypeInterface
      * $options.
      *
      * @param mixed $data
-     *
-     * @return string
      */
-    public function render(Field $field, $data, array $options);
+    public function render(Field $field, $data, array $options): string;
```
Zales0123, line 27: I believe indeed it is a BC-break :( I would skip that for now, unfortunately. There is a part of me that says "let's just document it and move on with the modern syntax" but I'm not sure... cc @lchrusciel I need a second opinion 🖖

lchrusciel: I must agree :/ Perhaps we can sooner than later just release GrudBundle 2.0

Roshyo: Removed. But I believe we can leave the extending classes with that return type, can we ?

Zales0123 (approved):

Zales0123:
Thanks, Stephane! :tada:

Issue #61, opened 2020-06-16
[GridBundle] Impossible to create new Grid Driver


Roshyo:
**Sylius version affected**: 1.x (I believe all of them, from the 1.0. 1.3 for sure)

**Description**  
In order to fetch entities from an API and then  display them in a grid, we have to create at least a new Driver and register it.
The thing is, when trying to register the new Driver in SyliusGrid, this exception is thrown :

The value "your\/driver" is not allowed for path "sylius_resource.drivers.1". Permissible values: "doctrine\/orm", "doctrine\/mongodb-odm", "doctrine\/phpcr-odm"

This is due to this in Sylius\Bundle\ResourceBundle\DependencyInjection\Configuration.php
```php
...
    /**
     * @param ArrayNodeDefinition $node
     */
    private function addDriversSection(ArrayNodeDefinition $node): void
    {
        $node
            ->children()
                ->arrayNode('drivers')
                    ->defaultValue([SyliusResourceBundle::DRIVER_DOCTRINE_ORM])
                    ->enumPrototype()->values(SyliusResourceBundle::getAvailableDrivers())->end()
                ->end()
            ->end()
        ;
    }
...
```

So the first step is (do not reproduce at home) to override the availableDrivers and add our (all done in vendor...)

Then, in Sylius\Bundle\ResourceBundle\AbstractResourceBundle, there is a switch/case to also mpodify and add our new value.
Next step (nearly last one) we also have to override the switch/case in Sylius\Bundle\ResourceBundle\DependencyInjection\Driver\DriverProvider

And to finish, we are warned that the XML file of our provider does not exist. So we create a new one in 
Sylius/resource-bundle/Resources/config/services/integration/your/driver.xml

**Steps to reproduce**  
Install Sylius (GridBundle at least) and try to add a new Provider

**Possible Solution**  
Be more permissive in this bundle configuration. It is just one step away from being able to fetch data in API and display them in a grid.

mbabker:
https://github.com/Sylius/SyliusGridBundle/issues/61#issuecomment-475543043 has it right, you just need a service with the right DI tag and you should be fine.  With that said...

After a question in Slack this morning, I'm starting to wonder what the value is of the drivers config node in the first place.

[...]

PR #227, opened 2022-03-17
Deprecate the drivers config node

mbabker:
As suggested in https://github.com/Sylius/SyliusGridBundle/issues/61#issuecomment-1066881862, this deprecates the "drivers" config node in favor of feature detection.  The node only allows values supported by the resource bundle and causes confusion when folks try to create custom drivers (which can be done with the `sylius.grid_driver` tag without adding it to that config node), so it really doesn't offer any benefit to existing.

Instead, Symfony's API for detecting features available at compile time is used.  If an app has both `doctrine/orm` and `doctrine/doctrine-bundle` installed, then the ORM and DBAL drivers will be automatically enabled.  If an app has both `doctrine/phpcr-odm` and `doctrine/phpcr-bundle` installed, the (deprecated) PHPCR-ODM driver will be automatically enabled.  The bundles are included in the dependency list because the driver services have dependencies to services created by the respective bundles, so the presence of the managers alone cannot fulfill the service requirements.

If accepted as is, the "drivers" config node is no longer used by the bundle at all.  I guess this could be a minor inconvenience if someone is using this bundle in an application that has both the ORM and PHPCR-ODM (and their bundles) installed and was using the config to disable one of the drivers, but you can accomplish the same with a compiler pass if needed.

On a side note, to have a full suite of test scenarios that actually makes sure these optional integrations stay optional, the CI would really need to run the tests without each of the Doctrine dependencies installed (similar to the steps in the resource bundle CI), but that's going to be a much larger undertaking (especially for the deprecated PHPCR-ODM integration).

### DIFF
--- a/src/Bundle/DependencyInjection/Configuration.php
+++ b/src/Bundle/DependencyInjection/Configuration.php
@@ -40,6 +40,7 @@ final class Configuration implements ConfigurationInterface
         $node
             ->children()
                 ->arrayNode('drivers')
+                    ->setDeprecated('sylius/grid-bundle', '1.12', 'The child node "%node%" at path "%path%" is deprecated and will be removed in SyliusGridBundle 2.0. As of SyliusGridBundle 1.12, it is no longer used.')
                     ->defaultValue([SyliusGridBundle::DRIVER_DOCTRINE_ORM])
                     ->enumPrototype()->values(SyliusGridBundle::getAvailableDrivers())->end()
                 ->end()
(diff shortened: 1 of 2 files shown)


mbabker:
To do that adds a lot of overhead for minimal benefit IMO (as you now have to create a new service class, inject it to the DI extension, test that in isolation, rewrite the functional tests for the class to mock all 4 scenarios needed for full coverage, and then maintain it all as a public API because that “decorator to mock one static call” service is going to be placed somewhere so it’s not being duplicated in several repos).  By design, it’s not an API to be mocked, it is making decisions based on the actual environment, which to me means those behaviors are best tested with something more complex than unit tests.

PR #229, opened 2022-03-21
[GridMaker] Use snake case for grid names

loic425:
Before:
```php
public static function getName(): string
{
    return 'app_adminuser';
}
```

After:
```php
public static function getName(): string
{
    return 'app_admin_user';
}
```


lchrusciel (approved):

src/Bundle/Resources/config/skeleton/Grid.tpl.php
```diff
@@ -26,7 +31,7 @@ public function __construct()
 
     public static function getName(): string
     {
-        return 'app_<?= strtolower($entity->getShortName()) ?>';
+        return 'app_<?= Str::asSnakeCase(($entity->getShortName())) ?>';
```
lchrusciel, line 34: I'm only wondering if we couldn't use Symfony stringify component here

lchrusciel:
Thanks, Loïc! :1st_place_medal:

PR #270, opened 2022-10-28
[WIP] fix: Make criteria work for php grids

Florian-Merle:
Filters in YAML configuration have the option to define a `default_value`. This is not the case for PHP configuration.

But, in the case of YAML, the `default_value` [is eventually used for the criteria in the filter definition](https://github.com/Sylius/SyliusGridBundle/blob/1.12/src/Component/Definition/ArrayToDefinitionConverter.php#L121).

Because of that, the `ArrayToDefinitionConverter` will only work for YAML grids (because PHP grids do not have a `default_value`, but a [`criteria` key in its configuration array](https://github.com/Sylius/SyliusGridBundle/blob/1.12/src/Bundle/Builder/Filter/Filter.php#L180)).

This pull request aims to solve this issue by renaming the `default_value` to `criteria` in the [`ArrayGridProvider`](https://github.com/Sylius/SyliusGridBundle/pull/270/files#diff-d7dc1791e7a24cdca838e3a770c2aaa70435a45a7d67aa2651c2170d5cd66a57R73) for YAML grids and changing the `ArrayToDefintionConverter` so that the `criteria` (not the `default_value`) key is used to set the criteria.

Note that the `criteria` is mixed in the Filter definition, so I changed the type of the `criteria` in the Filter builder to reflect that.

### DIFF
--- a/src/Bundle/Builder/Filter/Filter.php
+++ b/src/Bundle/Builder/Filter/Filter.php
@@ -29,7 +29,7 @@ final class Filter implements FilterInterface
 
     private array $formOptions = [];
 
-    private array $criteria = [];
+    private mixed $criteria = [];
 
     private function __construct(string $name, string $type)
     {
@@ -141,12 +141,12 @@ final class Filter implements FilterInterface
         return $this;
     }
 
-    public function getCriteria(): array
+    public function getCriteria(): mixed
     {
         return $this->criteria;
     }
 
-    public function setCriteria(array $criteria): FilterInterface
+    public function setCriteria(mixed $criteria): FilterInterface
     {
         $this->criteria = $criteria;
 
@@ -177,7 +177,7 @@ final class Filter implements FilterInterface
             $output['form_options'] = $this->formOptions;
         }
 
-        if (count($this->criteria) > 0) {
+        if ($this->criteria !== []) {
             $output['criteria'] = $this->criteria;
         }
 
--- a/src/Component/Provider/ArrayGridProvider.php
+++ b/src/Component/Provider/ArrayGridProvider.php
@@ -63,6 +63,24 @@ final class ArrayGridProvider implements GridProviderInterface
 
         $gridConfiguration = $this->gridConfigurationRemovalsHandler->handle($gridConfiguration);
 
+        if (isset($gridConfiguration['filters'])) {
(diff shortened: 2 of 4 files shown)


src/Bundle/Builder/Filter/Filter.php
```diff
@@ -29,7 +29,7 @@ final class Filter implements FilterInterface
 
     private array $formOptions = [];
 
-    private array $criteria = [];
+    private mixed $criteria = [];
```
loic425, line 32: oh... you're right, criteria in Definition Filter is mixed.
Normally this is not a bc change, I've to check

loic425: maybe null instead of []...

src/Bundle/Builder/Filter/Filter.php
```diff
@@ -177,7 +177,7 @@ public function toArray(): array
             $output['form_options'] = $this->formOptions;
         }
 
-        if (count($this->criteria) > 0) {
+        if ($this->criteria !== []) {
```
loic425, line 180: maybe null !== $criteria

src/Component/Provider/ArrayGridProvider.php
```diff
@@ -63,6 +63,24 @@ public function get(string $code): Grid
...
+            $gridConfiguration['filters'] = $this->handleFiltersDefaultValue($gridConfiguration['filters']);
+        }
+
         return $this->converter->convert($code, $gridConfiguration);
     }
+
+    private function handleFiltersDefaultValue(array $filters): array
+    {
+        foreach ($filters as &$filter) {
+            if (false === isset($filter['default_value'])) {
```
loic425, line 76:
```suggestion
            if (!isset($filter['default_value'])) {
```

loic425 (changes requested):

src/Bundle/Builder/Filter/Filter.php
loic425: that does not work...
https://3v4l.org/ijVvU

We have to define a new setDefaultValue and getDefaultValue instead I think.
And maybe that's better.

loic425:
@Florian-Merle What do you think of this implementation instead? https://github.com/Sylius/SyliusGridBundle/pull/274

PR #273, opened 2022-11-03
[WIP] Adding a migration tool for YAML grids

mamazu:
There has been an inofficial version of the migrator at https://github.com/mamazu/grid-config-converter . But it would make more sense to have it here and have it be officially supported by Sylius.


src/Bundle/Command/MigrateGrid.php
```diff
@@ -0,0 +1,55 @@
...
+final class MigrateGrid extends Command
+{
+    public function __construct(
+        private ConfigMigration $configMigration,
+    ) {
+    }
+
+    protected function configure(): void
+    {
+        $this->setName('sylius:grid:migrate');
```
loic425, line 33: maybe `sylius:migrate-grid`. @Zales0123 ?

src/Bundle/Migration/CodeOutputter.php
```diff
@@ -0,0 +1,92 @@
...
+class CodeOutputter extends Standard
+{
+    public function printCode(array $code): string
+    {
+        $code = $this->prettyPrint($code);
+
+        return <<<CODE
+<?php
+/**
+ * This code is generated by the config converter under https://github.com/mamazu/grid-config-converter
```
loic425: Not necessary now, what do you think?

src/Bundle/Migration/CommonConverterTrait.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        if (is_bool($value)) {
+            return new ConstFetch(new Name($value ? 'true' : 'false'));
+        }
+
+        if ($value === null) {
+            return new ConstFetch(new Name('null'));
+        }
+        if (is_int($value)) {
```
loic425:
```suggestion
        }
        
        if (is_int($value)) {
```

src/Bundle/Migration/CommonConverterTrait.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        if (is_string($value)) {
+            return new String_($value);
+        }
+
+        if (is_bool($value)) {
+            return new ConstFetch(new Name($value ? 'true' : 'false'));
+        }
+
+        if ($value === null) {
```
loic425:
```suggestion
        if (null === $value) {
```

src/Bundle/Migration/ConfigMigration.php
```diff
@@ -0,0 +1,154 @@
...
+    public function convertName(string $gridName): string
+    {
+        return ucfirst(preg_replace_callback('#_\w#', static fn ($a) => strtoupper($a[0][1]), $gridName));
+    }
+
+    public function getResourceClass(array $gridConfiguration): string
+    {
+        $resourceClass = $gridConfiguration['driver']['options']['class'] ?? 'To be replaced with the correct class.';
+
+        if (strpos($resourceClass, '%') !== false) {
```
loic425: use str_contains

src/Bundle/test/src/Migration/SyliusAdminOrder.php
```diff
@@ -0,0 +1,197 @@
...
+class SyliusAdminOrder extends AbstractGrid implements ResourceAwareGridInterface
+{
+    public static function getName(): string
+    {
+        return 'sylius_admin_order';
+    }
+
+    public function getResourceClass(): string
+    {
+        return 'Sylius\\Model\\Order';
```
loic425: maybe we can detect if its is a FQCN here.

src/Bundle/test/src/Migration/SyliusAdminOrder.php
```diff
@@ -0,0 +1,197 @@
...
+                ->setOptions([
+                    'format' => 'd-m-Y H:i:s',
+                ]),
+            )
+            ->addField(
+                TwigField::create('number', '@SyliusAdmin/Order/Grid/Field/number.html.twig')
+                ->setLabel('sylius.ui.number')
+                ->setPath('.')
+                ->setSortable(true)
+                ->setOptions([
```
loic425: use addOption instead, it's safer to prevent overriding defaultOptions (as you already have on TwigField). But maybe we have to fix this before...

src/Bundle/test/src/Migration/namespace_grid.yaml
```diff
@@ -0,0 +1,12 @@
...
+    grids:
+        foo_with_namespace:
+            driver:
+                name: 'doctrine/orm'
+                options:
+                    class: "%sylius.model.order.class%"
+                    pagination:
+                        fetch_join_collection: false
+                        use_output_walkers: false
+            removals: []
```
loic425, line 11: I think that is important to add a test with removals, even if this is brain new

loic425:
@mamazu Just a few review, it's a pretty big PR, but thank you for that huge and great work.

src/Bundle/Migration/ConfigMigration.php
mamazu: Done

src/Bundle/test/src/Migration/namespace_grid.yaml
mamazu: I haven't really found out how they would be structured in the yaml. Would they just be an array of fields and actions to remove? If so how would I know which is which?

loic425:
@mamazu I had a little discussion with Florian and I was thinking about it. what do you think about using grid definition from the container? The reason why I ask that is we can miss grid events that can change a grid definition. But I don't remember if this event is on runtime or not (I have to check that). But no offense, I see this PR is huge.

src/Bundle/test/src/Migration/namespace_grid.yaml
loic425: Yes indeed, they are not in the doc... we have to add it into the docs. it's just an array of field keys.

src/Bundle/test/src/Migration/SyliusAdminOrder.php
mamazu: Also done.

mamazu:
@loic425
**About the grid definition from container:** This shouldn't be a problem. Any array that has the same structure as the YAML can work in the migration tool. The question is if you have the grid events registered as a class somewhere you might need to unregister them when you bake the changes into the grid definition.

**About the size of the PR:** I completely agree, that this PR is too big. The problem is that if I want to split it up, then we either get a migrator that only does half of the things (eg. only generates the action section of the grid) or something that doesn't work at all.

And to be fair, I can also completely understand that it is really hard to understand what is happening in the code if you haven't built it yourself.

loic425:
Great, I see that we are in the same position. Maybe I should checkout this branch to understand better how it works :+1: I think I need to play with it a little to have a biggest picture of your code.

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+    ) {
+    }
+
+    /** @var array<string> */
+    private array $classesToUse = [];
+
+    /** @var array<Node> */
+    private array $classNodes = [];
+
+    private ?Node $namespace = null;
```
loic425:
```suggestion
    /** @var array<string> */
    private array $classesToUse = [];

    /** @var array<Node> */
    private array $classNodes = [];

    private ?Node $namespace = null;
    
    public function __construct(
        private CodeOutputter $codeOutputter,
    ) {
    }
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+                $implements,
+            );
+        }
+
+        $this->classNodes[] = new Class_(new Identifier($className), $classConfiguration);
+    }
+
+    public function generateGetResourceClassFunction(string $resourceNameOrString): Node
+    {
+        // If it's a class generate something like: Order::class
```
loic425:
```suggestion
        // If it's a class, it generates something like: Order::class
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+                ),
+                'stmts' => $body,
+            ],
+        );
+    }
+
+    /**
+     * @return array<Use_>
+     */
+    private function getUseStatementNodes(): array
```
loic425, line 149: this private method should be after the public ones

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+use PhpParser\Node\Name;
+use PhpParser\Node\Param;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
+use PhpParser\Node\Stmt\Use_;
+use PhpParser\Node\Stmt\UseUse;
+
+class CodeGenerator
```
loic425:
```suggestion
final class CodeGenerator
```

src/Bundle/Migration/CodeOutputter.php
```diff
@@ -0,0 +1,77 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\GridBundle\Migration;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Stmt;
+use PhpParser\PrettyPrinter\Standard;
+
+class CodeOutputter extends Standard
```
loic425:
```suggestion
final class CodeOutputter extends Standard
```

src/Bundle/Migration/ConfigMigration.php
```diff
@@ -0,0 +1,124 @@
...
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Stmt\Expression;
+use Sylius\Bundle\GridBundle\Builder\Filter\Filter;
+use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
+use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
+use Sylius\Bundle\GridBundle\Grid\ResourceAwareGridInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Yaml\Yaml;
+
+class ConfigMigration
```
loic425:
```suggestion
final class ConfigMigration
```

src/Bundle/Migration/GridBodyGenerator.php
```diff
@@ -0,0 +1,261 @@
...
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Name;
+use Sylius\Bundle\GridBundle\Builder\Field\DateTimeField;
+use Sylius\Bundle\GridBundle\Builder\Field\Field;
+use Sylius\Bundle\GridBundle\Builder\Field\StringField;
+use Sylius\Bundle\GridBundle\Builder\Field\TwigField;
+
+class GridBodyGenerator
```
loic425:
```suggestion
final class GridBodyGenerator
```

loic425 (changes requested):

loic425:
Something I have to purpose to you is to keep this huge work open as a PoC, create a branch named grid-converter in the grid repository (my side), and add a PR for each services you created one at a time (your side):
- ActionMethodGenerator
- CodeGenerator
- CodeOutputter 
- etc...

with Phpspec for each ones.
Are you ok with that?

mamazu:
That sounds like a good action plan. The only think I would like to add to that is:

If you want to have comprehensible tests I would recommend that we use PhpUnit. The reason behind that is that I would like to compare source code and not objects. Let me give you an example: If I test the correctness of the `createField` function that generates the code for a field. Then it would be much easier to have a test doing this:
```php
public function testCreateField() {
    $field = $this->fieldCreator->createField();

    $fieldCode = $this->printPhpCode($field);

    $this->assertEquals($fieldCode, 'TwigField::create("test", "test.html.twig")');
}
```
instead of
```php
public function it_creates_a_field() {
    $field = $this->createField();

    $field->shouldBe(new StaticMethod(TwigField::create, [new Arg(new String_('test'), new Arg(new String_('test.htl.twig'))]);
}
```

And this is just a simple example.

I don't know if you can do the upper example in PhpSpec as well. If so, then I will very gladly do that.

src/Bundle/Migration/CodeOutputter.php
mamazu: Has been removed.

loic425:
Yes it dépends on what do you test. Phpspec is for low level tests and phpunit for high level tests. On the grid builders, we have phpspec for grid objets like Field builders, but we have phpunit tests for grid configuration testing

mamazu:
Okay then I will do all the tests in PhpUnit and see what I can reasonably convert to spec tests.

PR #274, opened 2022-11-04
[GridBuilder] Add missing default value on filter

loic425:
Replace #270


Florian-Merle:
@loic425 I'm fine with this implementation.

However, IMO the criteria should probably be marked as deprecated because it has the same purpose the default_value has but for a YAML-configured grid.

lchrusciel (approved):

Roshyo (approved):

PR #275, opened 2022-11-06
Adding the code generator with tests

mamazu:
Grid Migration part 1

Related to #273

### DIFF
--- a/src/Bundle/Migration/CodeGenerator.php
+++ b/src/Bundle/Migration/CodeGenerator.php
@@ -67,7 +67,7 @@ final class CodeGenerator
         return new Name(end($namespaceParts));
     }
 
-    public function addClass(string $className, ?string $extends, ?array $implements, array $body): void
+    public function addClass(string $className, ?string $extends, array $implements = [], array $body = []): void
     {
         $classConfiguration = ['stmts' => $body];
 
@@ -75,7 +75,7 @@ final class CodeGenerator
             $classConfiguration['extends'] = new Identifier($extends);
         }
 
-        if (count($implements ?? []) > 0) {
+        if (count($implements) > 0) {
             $classConfiguration['implements'] = array_map(
                 static fn (string $interfaceName) => new Identifier($interfaceName),
                 $implements,
--- a/src/Bundle/Migration/CodeGenerator.php
+++ b/src/Bundle/Migration/CodeGenerator.php
@@ -46,6 +46,7 @@ final class CodeGenerator
         $this->namespace = new Node\Stmt\Namespace_(new Name($namespace));
     }
 
+    /** @param class-string $classToUse */
     public function addUseStatement(string $classToUse): void
     {
         $this->classesToUse[] = $classToUse;
@@ -67,11 +68,12 @@ final class CodeGenerator
         return new Name(end($namespaceParts));
     }
 
+    /** @param class-string $className */
     public function addClass(string $className, ?string $extends, array $implements = [], array $body = []): void
     {
         $classConfiguration = ['stmts' => $body];
 
-        if ($extends !== null) {
+        if (null !== $extends) {
             $classConfiguration['extends'] = new Identifier($extends);
         }
 
@@ -146,7 +148,7 @@ final class CodeGenerator
     public function build(): string
     {
         $nodes = [];
-        if ($this->namespace !== null) {
+        if (null !== $this->namespace) {
             $nodes[] = $this->namespace;
         }
 
--- a/src/Bundle/Migration/CodeGenerator.php
+++ b/src/Bundle/Migration/CodeGenerator.php
@@ -1,0 +1,183 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\GridBundle\Migration;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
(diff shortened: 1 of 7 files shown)


mamazu:
About the pipeline. I am not sure if we can ignore those errors. But I don't think fixing them is a good idea.

loic425:
Thx, I'll check that soon. Do not hesitate to ping me if I forgot.

src/Bundle/test/src/Migration/CodeGenatorTest.php
```diff
@@ -0,0 +1,200 @@
...
+
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Name;
+use PhpParser\Node\Scalar\String_;
+use PhpParser\Node\Stmt\Echo_;
+use PHPUnit\Framework\TestCase;
+use Sylius\Bundle\GridBundle\Migration\CodeGenerator;
+use Sylius\Bundle\GridBundle\Migration\CodeOutputter;
+
+class CodeGenatorTest extends TestCase
```
loic425:
```suggestion
final class CodeGeneratorTest extends TestCase
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+        private CodeOutputter $codeOutputter,
+    ) {
+    }
+
+    public function setNamespace(string $namespace): void
+    {
+        $this->namespace = new Node\Stmt\Namespace_(new Name($namespace));
+    }
+
+    public function addUseStatement(string $classToUse): void
```
loic425, line 50:
```suggestion
    /** @param class-string $classToUse */
    public function addUseStatement(string $classToUse): void
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+    public function getRelativeClassName(string $className): Name
+    {
+        $this->addUseStatement($className);
+
+        $namespaceParts = explode('\\', $className);
+
+        return new Name(end($namespaceParts));
+    }
+
+    public function addClass(string $className, ?string $extends, array $implements = [], array $body = []): void
```
loic425, line 72:
```suggestion
    @param class-string $className
    public function addClass(string $className, ?string $extends, array $implements = [], array $body = []): void
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+    {
+        $classConfiguration = ['stmts' => $body];
+
+        if ($extends !== null) {
```
loic425: Maybe use Yoda style

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+                $implements,
+            );
+        }
+
+        $this->classNodes[] = new Class_(new Identifier($className), $classConfiguration);
+    }
+
+    public function generateGetResourceClassFunction(string $resourceNameOrString): Node
+    {
+        // If it's a class generate something like: Order::class
```
loic425, line 92:
```suggestion
        // If it's a class, we need to generate something like: Order::class
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+                ),
+                'stmts' => $body,
+            ],
+        );
+    }
+
+    public function build(): string
+    {
+        $nodes = [];
+        if ($this->namespace !== null) {
```
loic425:
```suggestion
        if (null !== $this->namespace) {
```

src/Bundle/Migration/CodeGenerator.php
```diff
@@ -0,0 +1,183 @@
...
+            $nodes[] = $useStatements;
+        }
+
+        foreach ($this->classNodes as $classNode) {
+            $nodes[] = $classNode;
+        }
+
+        $generatedCode = "<?php\n" . $this->codeOutputter->prettyPrint($nodes);
+
+        $this->classNodes = [];
```
loic425, line 165: is it to reset values?

src/Bundle/test/src/Migration/CodeGenatorTest.php
```diff
@@ -0,0 +1,200 @@
...
+        $this->codeOutputter = new CodeOutputter();
+        $this->codeGenerator = new CodeGenerator($this->codeOutputter);
+    }
+
+    public function testGeneratingAnEmptyFile(): void
+    {
+        $this->assertEquals("<?php\n", $this->codeGenerator->build());
+    }
+
+    /** @dataProvider dataGeneratingImportStatements */
```
loic425: what does this annotation mean?

src/Bundle/Migration/CodeGenerator.php
mamazu: For this it would probably be the best to add a check in the codestyle of the plugin for that. But I can fix those as well.

src/Bundle/Migration/CodeGenerator.php
mamazu: Yes the code generator is stateful sadly. The reason for this is that we need a way to accumulate the list of import statements. In order to ensure that when the generator is used on multiple grids, we need to reset the code generator.

src/Bundle/test/src/Migration/CodeGenatorTest.php
mamazu: This is the [dataProvider Annotation](https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#data-providers). This means that the test has parameters and in the `dataGeneratingImportStatements` is generating the test cases. This is an easy way to stamp out tests which work are just the same execution flow but with different results.

mamazu:
> About the pipeline. I am not sure if we can ignore those errors. But I don't think fixing them is a good idea.

@loic425 Do you have any ideas on this one?

mamazu:
I don't think that it makes sense to have a migration tool in the grid bundle. It's a thing that you only run once and then be done with it.

PR #276, opened 2022-11-09
Autoconfigure grid filters

loic425:


src/Component/Filtering/AbstractFilter.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Filtering;
+
+abstract class AbstractFilter implements FilterInterface, TypeAwareFilterInterface, FormTypeAwareFilterInterface
```
lchrusciel: What would you say about the following change in the naming?
```suggestion
interface ConfigurableFilterInterface implements FilterInterface, TypeAwareFilterInterface, FormTypeAwareFilterInterface
```

lchrusciel (approved):
Great improvement 👍  But I don't see, why should we use abstract class for this new feature

src/Component/Filtering/AbstractFilter.php
loic425: Great Idea ! Indeed that's better than the abstract class...

lchrusciel (approved):

PR #298, opened 2023-03-09
Fix CI

loic425:


src/Bundle/test/src/Entity/Author.php
```diff
@@ -25,12 +25,14 @@ class Author implements ResourceInterface
 {
     /**
      * @Serializer\Expose
+     *
```
Rafikooo, line 28: I believe that these blanks can be ignored by ECS. We configured it earlier for behat and ORM, for instance, here:
https://github.com/Sylius/PriceHistoryPlugin/blob/main/ecs.php#L14-L17

Rafikooo (approved):

lchrusciel (approved):

PR #301, opened 2023-03-13
Get Grid by FQCN

loic425:
image

### DIFF
--- a/src/Bundle/Provider/ServiceGridProvider.php
+++ b/src/Bundle/Provider/ServiceGridProvider.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\Provider;
 
+use Sylius\Bundle\GridBundle\Grid\GridInterface;
 use Sylius\Bundle\GridBundle\Registry\GridRegistryInterface;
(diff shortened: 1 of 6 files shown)


Roshyo (approved):

Rafikooo (approved):

TheMilek (approved):

TheMilek:
Thank you, Loïc! :tada:

src/Bundle/Provider/ServiceGridProvider.php
```diff
@@ -47,6 +48,10 @@ public function __construct(
 
     public function get(string $code): Grid
     {
+        if (is_a($code, GridInterface::class, true)) {
+            $code = $code::getName();
+        }
```
lchrusciel, line 53: Am I the only one who seems to have a problem with the flow here? If the code is GridInterface, then code equals to the name of grid? 😂 Let's maybe at least rename it to an identifier 
```suggestion
        if (is_a($identifier, GridInterface::class, true)) {
            $identifier = $identifier::getName();
        }
````

or let's teach grid registry to look for grids by FQCN

loic425: Good Idea to rename it identifier

PR #307, opened 2023-06-19
Move tests app

loic425:


jakubtobiasz (approved):
Similarly to ResourceBundle, we can move the `spec` dir to the root directory too.

ecs.php
```diff
@@ -9,25 +9,28 @@
...
-        ->set(HeaderCommentFixer::class)->call('configure', [[
-            'location' => 'after_open',
-            'header' =>
-'This file is part of the Sylius package.
+    $ecsConfig->ruleWithConfiguration(HeaderCommentFixer::class, [
+        'location' => 'after_open',
+        'header' =>
+            'This file is part of the Sylius package.
 
 (c) Paweł Jędrzejewski
```
lchrusciel, line 25: @Rafikooo you should update licences in technical libs as well, shouldn't you?

lchrusciel (approved):

lchrusciel:
Thank you, @loic425!

ecs.php
loic425: yes, Lukasz, it will be another PR.

Rafikooo: @lchrusciel, sure! 🚀

PR #312, opened 2023-10-13
:sparkles: Allow join on ExpressionBuilder comparison

FabienCharlier:
Allowing for the compare method of ExpressionBuilder to join on different table before comparing.

It is currently not possible to compare two fields which are in a joined table.

### DIFF
--- a/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
+++ b/src/Bundle/Doctrine/ORM/ExpressionBuilder.php
@@ -40,7 +40,10 @@ final class ExpressionBuilder implements MemberOfAwareExpressionBuilderInterface
 
     public function comparison(string $field, string $operator, $value)
     {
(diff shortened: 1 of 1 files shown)


lchrusciel:
Hey Fabien! 

Thanks a lot for your contribution, but without a functional test, we won't merge this PR. Could you add it?

FabienCharlier:
I'd like to, but I'm really not sure how to do it. There isn't a lot of examples in the Functional test folder, and nothing looks really close to what I'm trying to do.
To setup a functionnal test, I'd ideally want to have a database with two tables (one referencing another with a foreign key), a grid setup on one of those tables. This grid would need to have a custom filter, which tries to compare two fields from the joined table. Then, requiring the correct url should do the trick.

Do you have any documentation or similar test cases that I could read to help me create a test case ?

PR #321, opened 2024-03-22
Add callable field

Florian-Merle:
> The Callable column aims to offer almost as much flexibility as the Twig column, but without requiring the creation of a template.
> You simply need to specify a callable, which allows you to transform the 'data' variable on the fly.

The documentation associated for this PR can be found here https://github.com/Sylius/Stack/pull/228


docs/field_types.md
```diff
@@ -312,3 +312,89 @@ $field->setOptions([
     // Your options here
 ]);
 ```
+
+Callback
+------
```
loic425:
```suggestion
Callback
--------
```

docs/field_types.md
```diff
@@ -312,3 +312,89 @@ $field->setOptions([
...
+
+The callback column aims to offer almost as much flexibility as the twig column, but without requiring the creation of a template.
```
loic425:
```suggestion
The callback column aims to offer almost as much flexibility as the Twig column, but without requiring the creation of a template.
```

loic425 (approved):

src/Bundle/Resources/config/services/field_types.xml
```diff
@@ -15,6 +15,12 @@
     <services>
         <defaults public="true" />
 
+        <service id="Sylius\Component\Grid\FieldTypes\CallbackFieldType">
+            <argument type="service" id="sylius.grid.data_extractor" />
+            <tag name="sylius.grid_field" type="callback" />
+        </service>
+        <service id="sylius.grid_field.string" alias="Sylius\Component\Grid\FieldTypes\StringFieldType" />
```
diimpp:
```suggestion
        <service id="sylius.grid_field.callback" alias="Sylius\Component\Grid\FieldTypes\CallbackFieldType" />
```

src/Component/FieldTypes/CallbackFieldType.php
```diff
@@ -0,0 +1,49 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class CallbackFieldType implements FieldTypeInterface
+{
+    private DataExtractorInterface $dataExtractor;
+
+    public function __construct(DataExtractorInterface $dataExtractor)
+    {
+        $this->dataExtractor = $dataExtractor;
+    }
```
diimpp:
```suggestion
    public function __construct(private DataExtractorInterface $dataExtractor)
    {
    }
```

src/Component/FieldTypes/CallbackFieldType.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $value = $this->dataExtractor->get($field, $data);
+        $value = call_user_func($options['callback'], $value);
+
+        if ($options['htmlspecialchars'] !== true) {
+            return (string) $value;
+        }
```
diimpp: Just a suggestion, looks a bit unusual to have `!== true`on bool typed property.

```suggestion
        if ($options['htmlspecialchars'] === false) {
            return (string) $value;
        }
```
```suggestion
        if (!$options['htmlspecialchars']) {
            return (string) $value;
        }
```

src/Component/FieldTypes/CallbackFieldType.php
```diff
@@ -0,0 +1,49 @@
...
+
+    public function __construct(DataExtractorInterface $dataExtractor)
+    {
+        $this->dataExtractor = $dataExtractor;
+    }
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $value = $this->dataExtractor->get($field, $data);
+        $value = call_user_func($options['callback'], $value);
```
diimpp: You can typecast over here instead of doing it twice at both returns down bellow. 

```suggestion
        $value = (string) call_user_func($options['callback'], $value);
```

src/Component/spec/FieldTypes/CallbackFieldTypeSpec.php
```diff
@@ -0,0 +1,85 @@
...
+    ): void {
+        $dataExtractor->get($field, ['foo' => 'bar'])->willReturn('BAR');
+
+        $this->render($field, ['foo' => 'bar'], [
+            'callback' => [self::class, 'callable'],
+            'htmlspecialchars' => true,
+        ])->shouldReturn('bar');
+    }
+
+    static function callable($value)
```
diimpp: Should be possible to typecast it?

```suggestion
    static function callable(mixed $value): string
```

docs/field_types.md
```diff
@@ -312,3 +312,89 @@ $field->setOptions([
...
+Callback
+--------
+
+The Callback column aims to offer almost as much flexibility as the Twig column, but without requiring the creation of a template.
+You simply need to specify a callback, which allows you to transform the 'data' variable on the fly.
+
+By default it uses the name of the field, but you can specify the path
+alternatively. For example:
+
+<details open><summary>PHP</summary>
```
diimpp: To match existing documentation, please add YAML example as well.

diimpp (changes requested):

docs/field_types.md
Florian-Merle: I did not include documentation for the YAML format because I believe it's impossible to define a callback field in this format.

diimpp: Feature is not complete then and cannot be merged until YAML support will be dropped.

Here is example on how callback can be implemented with non-php config format.

https://symfony.com/doc/current/reference/constraints/Callback.html#external-callbacks-and-closures

loic425: I think we can do it. The repository can be a callback in YAML.
```yaml
app_book_by_english_authors:
    driver:
        name: doctrine/orm
        options:
            class: App\Entity\Book
            repository:
                method: [expr:service('app.english_books_query_builder'), create]
```

Florian-Merle: I reworked the feature and introduced a new way to define callbacks, inspired by Loic's comment:

```yaml
sylius_grid:
    grids:
        app_user:
            fields:
                name:
                    type: callback
                    options:
                        callback: "callback:strtoupper"
                    label: app.ui.name
```

This approach allows string callables to be defined directly in YAML.
When configuring grids using PHP, both string and array callables are supported.
When configuring grids using service definitions, everything is supported: string callables, array callables, and closures.

I’ve documented this change here: [SyliusGridBundle PR #321.](https://github.com/Sylius/SyliusGridBundle/pull/321/files#diff-9d9f0cce19acf39b9271383ce30906ed2c529f88929259db65a730d247fea400R322)

To implement this, I copied the `OptionParser` class from the ResourceBundle, slimming it down to only what’s necessary for this feature (handling callbacks). See the original class: [OptionsParser](https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Bundle/Grid/Parser/OptionsParser.php) in ResourceBundle.

Since the resource bundle is not a required dependencie, I think this is fine to copy the class. However having two `OptionParser` might not be ideal. Please let me know what you think about it.

Florian-Merle:
~~After some thinking, I believe it would be more appropriate to rename `CallbackField` into `CallableField`. I'll refactor that.~~ -> this has been done

src/Bundle/Resources/config/services.xml
```diff
@@ -128,5 +128,8 @@
             <tag name="maker.command" />
         </service>
         <service id="Sylius\Bundle\GridBundle\Maker\MakeGrid" alias="sylius.grid.maker" />
+
+        <service id="sylius.grid.options_parser" class="Sylius\Bundle\GridBundle\Parser\OptionsParser" />
```
Florian-Merle: @GSadee Since the SyliusResourceBundle has already a service with a similar id (`sylius.grid_options_parser`), i'm not sure about what this service should be named. WDYT ?

loic425: I think it should be
`sylius_grid.options_parser`

loic425: @Florian-Merle I've just found the ADR for the naming convention.
https://github.com/Sylius/Sylius/blob/2.0/adr/2024_10_03_services_naming_convention.md

But it doesn't care about the real naming.
However, in the example, we have "sylius_admin" and I think we/they decided to suffix with the pakage name.

docs/field_types.md
```diff
@@ -319,3 +319,118 @@ $field->setOptions([
     // Your options here
 ]);
 ```
+
+Callable
+--------
+
+The Callable column aims to offer almost as much flexibility as the Twig column, but without requiring the creation of a template.
```
loic425: For the documentation, please open a PR on the Sylius stack documentation.

Florian-Merle: Done, see https://github.com/Sylius/Stack/pull/228

src/Bundle/Parser/OptionsParser.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    public function parseOptions(array $parameters): array
+    {
+        return array_map(
+            /**
+             * @param mixed $parameter
+             *
+             * @return mixed
+             */
+            function ($parameter) {
```
loic425:
```suggestion
            function (mixed $parameter): mixed {
```

src/Bundle/Parser/OptionsParser.php
```diff
@@ -0,0 +1,63 @@
...
+            $parameters,
+        );
+    }
+
+    /**
+     * @param mixed $parameter
+     *
+     * @return mixed
+     */
+    private function parseOption($parameter)
```
loic425:
```suggestion
    private function parseOption(mixed $parameter): mixed
```

src/Bundle/Resources/config/services.xml
loic425:
```suggestion
        <service id="sylius_grid.options_parser" class="Sylius\Bundle\GridBundle\Parser\OptionsParser" />
```

src/Bundle/Resources/config/services/field_types.xml
```diff
@@ -15,6 +15,12 @@
     <services>
         <defaults public="true" />
 
+        <service id="sylius.grid_field.callable" class="Sylius\Component\Grid\FieldTypes\CallableFieldType">
```
loic425, line 18:
```suggestion
        <service id="sylius_grid.field.callable" class="Sylius\Component\Grid\FieldTypes\CallableFieldType">
```

src/Component/FieldTypes/CallableFieldType.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $value = $this->dataExtractor->get($field, $data);
+        $value = call_user_func($options['callable'], $value);
+
+        try {
+            $value = (string) $value;
+        } catch (\Throwable $e) {
```
loic425, line 34:
```suggestion
        } catch (\Throwable) {
```

src/Bundle/Resources/config/services.xml
Florian-Merle: I guess it makes more sense :)

docs/field_types.md
loic425: So, you should revert the doc part from this PR ;)

Florian-Merle: My bad, I thought that since the doc is still on the `1.14` branch, the doc should still be updated

src/Bundle/Parser/OptionsParser.php
```diff
@@ -0,0 +1,53 @@
...
+        );
+    }
+
+    private function parseOption(mixed $parameter): mixed
+    {
+        if (!is_string($parameter)) {
+            return $parameter;
+        }
+
+        if (0 === strpos($parameter, 'callable:')) {
```
loic425:
```suggestion
        if (str_starts_with($parameter, 'callable:')) {
```

phpstan.neon
```diff
@@ -1,4 +1,6 @@
 includes:
+    - phpstan-baseline.neon
+
```
diimpp: Let's remove whiteline between baseline and vendor includes.

```suggestion
```

Other main packages don't do it.

https://github.com/Sylius/Stack/blob/c28e70ca7c2cf48422212ca1278dc48d9ed56f89/phpstan.neon#L2
https://github.com/Sylius/Sylius/blob/4f0ab4bb43a9275f1b19170d83ea18c3cb3c1613/phpstan.neon.dist#L2

src/Bundle/Resources/config/services.xml
```diff
@@ -128,5 +128,8 @@
             <tag name="maker.command" />
         </service>
         <service id="Sylius\Bundle\GridBundle\Maker\MakeGrid" alias="sylius.grid.maker" />
+
+        <service id="sylius_grid.options_parser" class="Sylius\Bundle\GridBundle\Parser\OptionsParser" />
```
diimpp: This whole service file is public by default, hence we need to un-public it explicitly. 

```suggestion
        <service id="sylius_grid.options_parser" class="Sylius\Bundle\GridBundle\Parser\OptionsParser" public="false" />
```

diimpp: While @loic425 mentions an ADR in https://github.com/Sylius/SyliusGridBundle/pull/321#discussion_r1930124957, this whole file follows `sylius.grid.***` scheme and I think we shouldn't introduce `sylius_grid` just for this service alone.

Hence
```suggestion
        <service id="sylius.grid.options_parser" class="Sylius\Bundle\GridBundle\Parser\OptionsParser" />
```

P.S. It doesn't conflict with resource's option parser service. https://github.com/Sylius/SyliusResourceBundle/blob/0228f7593b63c9cb77edf4f8cae4f235bb73d37e/src/Bundle/Resources/config/services/integrations/grid.xml#L48

src/Bundle/Resources/config/services.xml
```diff
@@ -128,5 +128,8 @@
...
+        <service id="Sylius\Bundle\GridBundle\Parser\OptionsParserInterface" alias="sylius_grid.options_parser" />
```
diimpp:
```suggestion
        <service id="Sylius\Bundle\GridBundle\Parser\OptionsParserInterface" alias="sylius.grid.options_parser" />
```

src/Bundle/Resources/config/services/twig.xml
```diff
@@ -23,6 +23,7 @@
             <argument>@SyliusGrid/_grid.html.twig</argument>
             <argument>%sylius.grid.templates.action%</argument>
             <argument>%sylius.grid.templates.filter%</argument>
+            <argument type="service" id="sylius_grid.options_parser" />
```
diimpp:
```suggestion
            <argument type="service" id="sylius.grid.options_parser" />
```

src/Bundle/Tests/Unit/Parser/OptionsParserTest.php
```diff
@@ -0,0 +1,56 @@
...
+use Sylius\Bundle\GridBundle\Parser\OptionsParserInterface;
+
+final class OptionsParserTest extends TestCase
+{
+    public function testItImplementsOptionsParserInterface(): void
+    {
+        $this->assertInstanceOf(OptionsParserInterface::class, new OptionsParser());
+    }
+
+    public function testItParserOptionsWithCallable(): void
```
diimpp:
```suggestion
    public function testItParsesOptionsWithCallable(): void
```

src/Component/FieldTypes/CallableFieldType.php
```diff
@@ -0,0 +1,52 @@
...
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $value = $this->dataExtractor->get($field, $data);
+        $value = call_user_func($options['callable'], $value);
+
+        try {
+            $value = (string) $value;
+        } catch (\Throwable) {
+            throw new \RuntimeException(\sprintf('The callback for field "%s" returned a value that could not be converted to string.', $field->getName()));
```
diimpp: How about this?

Symfony uses this new syntax in the docs quite a lot
> Object (id 1) is not found.


```suggestion
            throw new \RuntimeException(\sprintf('Callable field (name "%s") returned value could not be converted to string.', $field->getName()));
```

Additionally we can either show actual error via `\Throwable $e->getMessage()` and get type of returned value via `get_debug_type()`

src/Component/spec/FieldTypes/CallableFieldTypeSpec.php
```diff
@@ -0,0 +1,104 @@
...
+    function it_uses_data_extractor_to_obtain_data_and_passes_it_to_a_static_callable(
+        DataExtractorInterface $dataExtractor,
+        Field $field,
+    ): void {
+        $dataExtractor->get($field, ['foo' => 'bar'])->willReturn('BAR');
+
+        $this->render($field, ['foo' => 'bar'], [
+            'callable' => [self::class, 'callable'],
+            'htmlspecialchars' => true,
+        ])->shouldReturn('bar');
```
diimpp: Please make sure input and output values are actually affected by the function to avoid case when callable is not applied somehow.


```suggestion
        $this->render($field, ['foo' => 'BAR'], [
            'callable' => [self::class, 'callable'],
            'htmlspecialchars' => true,
        ])->shouldReturn('bar');
```

tests/Application/config/sylius/grids/author.php
```diff
@@ -22,9 +23,8 @@
             ->addFilter(StringFilter::create('name'))
             ->orderBy('name', 'asc')
             ->addField(
-                StringField::create('id')
-                    ->setSortable(true)
-                    ->setEnabled(false),
```
diimpp, line 27: Wasn't `enabled: false` part of some other test?

src/Bundle/Parser/OptionsParser.php
```diff
@@ -0,0 +1,53 @@
...
+            return $this->parseOptionCallable(substr($parameter, 9));
+        }
+
+        return $parameter;
+    }
+
+    private function parseOptionCallable(string $callable): \Closure
+    {
+        if (!is_callable($callable)) {
+            throw new \RuntimeException(\sprintf('%s is not a callable.', $callable));
```
diimpp:
```suggestion
            throw new \InvalidArgumentException(sprintf('%s is not a callable.', $callable));
```

@loic425 I wonder if should introduce 
> Exception\ExceptionInterface
> Exception\InvalidArgumentException

exceptions in this PR to avoid potentional BC in the future.

src/Component/FieldTypes/CallableFieldType.php
diimpp: I wonder if it should be a named exception, like `FieldTypeReturnTypeStringCastException` or something along the lines.

tests/Application/config/sylius/grids/author.php
Florian-Merle: Changing it did not break the tests, so I believe it's not part of other tests. I took the liberty of modifying this field to make it easier for me to write my tests. If needed, I can always add another field for my tests. Let me know

src/Component/FieldTypes/CallableFieldType.php
Florian-Merle: Reading the description of [`UnexpectedValueException`](https://www.php.net/manual/en/class.unexpectedvalueexception.php), I believe this is the appropriate exception to use. I updated the code accordingly. Let me know if this is fine with you.

diimpp (approved):
@Florian-Merle :+1:

GSadee (approved):

GSadee:
Thank you @Florian-Merle! 🎉

PR #327, opened 2024-06-04
Test grids with UI instead of API

loic425:


src/Bundle/Tests/Functional/GridUiTest.php
```diff
@@ -13,9 +13,13 @@
...
+use Coduo\PHPMatcher\Backtrace\VoidBacktrace;
+use Coduo\PHPMatcher\Matcher;
+use Symfony\Component\DomCrawler\Crawler;
+use Symfony\Component\HttpFoundation\Response;
 
-final class GridApiTest extends JsonApiTestCase
+final class GridUiTest extends ApiTestCase
 {
     /** @var array */
     private $data;
```
diimpp:
```suggestion
    private array $data;
```

tests/Application/config/routes.yaml
```diff
@@ -3,13 +3,21 @@ app_books:
         alias: app.book
         grid: app_book
         only: ['index']
+        templates: crud
```
diimpp, line 6: Does it needs `templates` config when it's a `sylius.resource_api`? Same for other routes in this file.

tests/Application/src/Grid/AuthorWithBooksWithUseOutputWalkersEnabled.php
```diff
@@ -38,7 +38,8 @@ public function buildGrid(GridBuilderInterface $gridBuilder): void
             ->setRepositoryMethod(["expr:service('app.authors_with_books_query_builder')", 'create'])
             ->addField(
                 StringField::create('book')
-                ->setSortable(true, 'book.title'),
+                    ->setPath('books[0].title')
+                    ->setSortable(true, 'book.title'),
```
diimpp, line 42: > ->setPath('books[0].title')

Feels weird, but I don't have any points.

src/Bundle/Tests/Functional/GridUiTest.php
```diff
@@ -214,40 +226,72 @@ public function it_filters_books_by_author_when_an_author_is_used_in_join_in_que
...
-        $totalItemsCountAfterSorting = $this->getTotalItemsCountFromCurrentResponse();
+
+        $totalItemsCountAfterSorting = count($this->getAuthorNames());
 
         $this->assertSame($totalItemsCountBeforeSorting, $totalItemsCountAfterSorting);
     }
 
-    private function getTotalItemsCountFromCurrentResponse(): int
+    /** @return string[] */
+    private function getBookTitles(): array
```
diimpp: Would be cool to indicate, that it's related to current response/response crawler as it was before.

From cases like this, just be reading code

```php
        $totalItemsCountBeforeSorting = count($this->getAuthorNames());
        $this->client->request('GET', '/authors/?sorting[nationality]=desc');
        $totalItemsCountAfterSorting = count($this->getAuthorNames());
```

it's not obvious, that `getAuthorNames` will return different result.

diimpp (approved):

src/Bundle/Tests/Functional/GridUiTest.php
loic425: I rename it getAuthorNamesFromResponse, maybe that's better?

tests/Application/config/routes.yaml
loic425: yep, I forgot to transform sylius.resource_api to sylius.resource, cause we do not test the api part anymore. It's already tested with resource bundle in the package. Cause hateoas does not support attribute instead of Doctrine annotations, we'll not able to test the grid bundle with symfony 7, if we do not transform all these tests with a frontend UI.

tests/Application/src/Grid/AuthorWithBooksWithUseOutputWalkersEnabled.php
loic425: yeah, not very cool but enough for the test itself.

diimpp (approved):

GSadee (approved):

GSadee:
Thank you, Loïc! :tada:

PR #338, opened 2024-07-19
Create action and filter context factories

loic425:
In Sylius resource bundle, a Twig Grid renderer decorates the grid one in order to add some data to the context (Twig context here but the renderer does not depends on Twig).

Cause I need to add things on resource side to this context, I think it will be easier to allow decorating that context instead.

Example for the future in Resource package to improve that https://github.com/Sylius/SyliusResourceBundle/blob/1.11/src/Bundle/Grid/Renderer/TwigGridRenderer.php#L64

```
final class ActionContextFactory implemented ActionContextFactoryInterface
{
    public function __construct(
        private ActionContextFactoryInterface $contextFactory,
        private OptionsParserInterface $optionsParser,
    ) {
    }

    public function create(GridViewInterface $gridView, Action $action, mixed $data = null): array
    {
        $options = $this->optionsParser->parseOptions(
            $action->getOptions(),
            $gridView->getRequestConfiguration()->getRequest(),
            $data,
        );

        return array_merge($this->contextFactory->create($grid, $action, $data), [
            'grid' => $gridView,
            'action' => $action,
            'data' => $data,
            'options' => $options,
        ]);
    }
}
```

Cause I will also need to add new context data when using the new routing system.

### DIFF
--- a/src/Bundle/Resources/config/services.xml
+++ b/src/Bundle/Resources/config/services.xml
@@ -15,6 +15,7 @@
     <imports>
         <import resource="services/field_types.xml" />
         <import resource="services/filters.xml" />
+        <import resource="services/renderer_context.xml" />
         <import resource="services/templating.xml" />
         <import resource="services/twig.xml" />
     </imports>
@@ -71,6 +72,8 @@
             <argument type="service" id="sylius.grid.data_provider" />
         </service>
         <service id="sylius.grid.view_factory" alias="Sylius\Component\Grid\View\GridViewFactoryInterface" />
+
+
 
         <service id="sylius.grid.data_provider" class="Sylius\Component\Grid\Data\DataProvider">
             <argument type="service" id="sylius.grid.data_source_provider" />
--- a/src/Bundle/Resources/config/services/renderer_context.xml
+++ b/src/Bundle/Resources/config/services/renderer_context.xml
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+
+ This file is part of the Sylius package.
+
+ (c) Sylius Sp. z o.o.
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <defaults public="false" />
+
+        <service id="sylius.grid.renderer.context.factory.action" class="Sylius\Component\Grid\Renderer\Context\Factory\ActionContextFactory" />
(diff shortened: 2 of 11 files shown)


src/Bundle/Resources/config/services.xml
```diff
@@ -72,6 +73,8 @@
         </service>
         <service id="sylius.grid.view_factory" alias="Sylius\Component\Grid\View\GridViewFactoryInterface" />
 
+
+
```
diimpp, line 77:
```suggestion
```

src/Bundle/Resources/config/services.xml
```diff
@@ -15,6 +15,7 @@
     <imports>
         <import resource="services/field_types.xml" />
         <import resource="services/filters.xml" />
+        <import resource="services/renderer_context.xml" />
         <import resource="services/templating.xml" />
         <import resource="services/twig.xml" />
```
diimpp, line 20: Just an idea, this approach is less error prone, less code and seems what sylius uses in newer bundles, like ApiBundle.
```suggestion
        <import resource="services/*.xml" />
```
```suggestion
        <import resource="services/**/*.xml" />
```

https://github.com/Sylius/Sylius/blob/b7d92a66caed63094d820296d20d00a1bee0ed04/src/Sylius/Bundle/ApiBundle/Resources/config/services.xml#L18-L20

src/Bundle/Resources/config/services/renderer_context.xml
```diff
@@ -0,0 +1,25 @@
...
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <defaults public="false" />
+
+        <service id="sylius.grid.renderer.context.factory.action" class="Sylius\Component\Grid\Renderer\Context\Factory\ActionContextFactory" />
+        <service id="Sylius\Component\Grid\Renderer\Context\Factory\ActionContextFactoryInterface" alias="sylius.grid.renderer.context.factory.action" />
+
+        <service id="sylius.grid.renderer.context.factory.filter" class="Sylius\Component\Grid\Renderer\Context\Factory\FilterContextFactory" />
+        <service id="Sylius\Component\Grid\Renderer\Context\Factory\FilterContextFactoryInterface" alias="sylius.grid.renderer.context.factory.filter" />
+
```
diimpp, line 23:
```suggestion
```

diimpp (commented):
(I'm just thinking out load, it's not a call to action.)

It's a bit strange to have `Context/Factory/ActionContextFactoryInterface`, but no actual `ActionContext` as "context" is array and no factoring is happening.

It's more like formatter or provider.

Is there any point adding parameter bag as in TwigHooks for that?

loic425:
> (I'm just thinking out load, it's not a call to action.)
> 
> It's a bit strange to have `Context/Factory/ActionContextFactoryInterface`, but no actual `ActionContext` as "context" is array and no factoring is happening.
> 
> It's more like formatter or provider.

Interesting solution here https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Serializer/Context/ContextBuilderInterface.php

PR #342, opened 2024-09-25
Add storage for filters

loic425:
This goal is to extract that https://github.com/Sylius/Sylius/blob/2.0/src/Sylius/Bundle/UiBundle/Storage/FilterStorage.php from UiBundle and be able use it in the new Sylius stack.

We need this one for "RedirectPathExtension" (from Ui bundle to Twig Extra package).

Then we'll need to extract AdminFilterSubscriber from AdminBundle (to move into AdminUi), but we can do this later.

### DIFF
--- a/src/Bundle/Resources/config/services/filters.xml
+++ b/src/Bundle/Resources/config/services/filters.xml
@@ -15,6 +15,12 @@
     <services>
         <defaults public="true" />
 
+        <service id="sylius_grid.symfony.session.filter_storage" class="Sylius\Component\Grid\Symfony\Session\Filter\Storage\SessionFilterStorage" public="false">
+            <argument type="service" id="request_stack" />
+        </service>
+        <service id="sylius_grid.filter_storage" alias="sylius_grid.symfony.session.filter_storage" />
+        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius_grid.filter_storage" />
+
         <service id="sylius.form_registry.grid_filter" class="Sylius\Bundle\GridBundle\Form\Registry\FormTypeRegistry" public="false" />
 
         <service id="Sylius\Component\Grid\Filter\StringFilter">
--- a/src/Bundle/Resources/config/services/filters.xml
+++ b/src/Bundle/Resources/config/services/filters.xml
@@ -19,7 +19,7 @@
             <argument type="service" id="request_stack" />
         </service>
         <service id="sylius.grid.filter_storage" alias="sylius.grid.filter_storage.session" />
-        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius.grid.filter_storage" />
+        <service id="Sylius\Bundle\GridBundle\Storage\FilterStorageInterface" alias="sylius.grid.filter_storage" />
 
         <service id="sylius.form_registry.grid_filter" class="Sylius\Bundle\GridBundle\Form\Registry\FormTypeRegistry" public="false" />
 
--- a/src/Bundle/Resources/config/services/filters.xml
+++ b/src/Bundle/Resources/config/services/filters.xml
@@ -15,11 +15,11 @@
     <services>
         <defaults public="true" />
 
-        <service id="sylius_grid.symfony.session.filter_storage" class="Sylius\Component\Grid\Symfony\Session\Filter\Storage\SessionFilterStorage" public="false">
+        <service id="sylius.grid.filter_storage.session" class="Sylius\Bundle\GridBundle\Storage\SessionFilterStorage" public="false">
             <argument type="service" id="request_stack" />
         </service>
-        <service id="sylius_grid.filter_storage" alias="sylius_grid.symfony.session.filter_storage" />
-        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius_grid.filter_storage" />
+        <service id="sylius.grid.filter_storage" alias="sylius.grid.filter_storage.session" />
+        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius.grid.filter_storage" />
 
         <service id="sylius.form_registry.grid_filter" class="Sylius\Bundle\GridBundle\Form\Registry\FormTypeRegistry" public="false" />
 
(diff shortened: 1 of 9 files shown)


src/Component/Symfony/Session/Filter/Storage/SessionFilterStorage.php
diimpp, line 1: I think we should keep this file at regular place at existing bundle, since grid package is not being upgraded to bundless file structure atm and it's better to upkeep to the same style.

src/Component/Symfony/Session/Filter/Storage/SessionFilterStorage.php
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Symfony\Session\Filter\Storage;
```
diimpp:
```suggestion
namespace Sylius\Bundle\GridBundle\Storage;
```

src/Bundle/Resources/config/services/filters.xml
```diff
@@ -15,6 +15,12 @@
     <services>
         <defaults public="true" />
 
+        <service id="sylius_grid.symfony.session.filter_storage" class="Sylius\Component\Grid\Symfony\Session\Filter\Storage\SessionFilterStorage" public="false">
+            <argument type="service" id="request_stack" />
+        </service>
+        <service id="sylius_grid.filter_storage" alias="sylius_grid.symfony.session.filter_storage" />
+        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius_grid.filter_storage" />
```
diimpp:
```suggestion
        <service id="sylius.grid.filter_storage.session" class="Sylius\Bundle\GridBundle\Storage\SessionFilterStorage" public="false">
            <argument type="service" id="request_stack" />
        </service>
        <service id="sylius.grid.filter_storage" alias="sylius.grid.filter_storage.session" />
        <service id="Sylius\Component\Grid\Filter\Storage\FilterStorageInterface" alias="sylius.grid.filter_storage" />
```

src/Component/Tests/Integration/Filter/Storage/SessionFilterStorageTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Tests\Integration\Filter\Storage;
+
+use Sylius\Component\Grid\Filter\Storage\FilterStorageInterface;
+use Sylius\Component\Grid\Symfony\Session\Filter\Storage\SessionFilterStorage;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+final class SessionFilterStorageTest extends KernelTestCase
```
diimpp: Does it mean each DI service in the project needs to be tested eventually? :eyes:

src/Component/Symfony/Session/Filter/Storage/SessionFilterStorage.php
diimpp: Commented on the wrong file -- I've meant `src/Component/Symfony/Session/Filter/Storage/SessionFilterStorage.php`

loic425: Hum. You are probably right. I'll have to do an alias after removing component namespace.

src/Component/Tests/Integration/Filter/Storage/SessionFilterStorageTest.php
loic425: Just testing the service configuration is minimalist, but testing with DI is interesting. Cause you will be able to test it with the session in real condition.

src/Component/Symfony/Session/Filter/Storage/SessionFilterStorage.php
loic425: What do you think about Sylius\Bundle\GridBundle\Session\Filter\Storage instead not to forget about the new namespace when moving to one-package architecture?

diimpp: It's still doesn't match existing notation, so generally I'm against it. As source code reader, it leaves bad impression, when codebase uses multiple structuring techniques. 

It's only a one file, so it shouldn't affect future one-package initiative.

src/Component/Tests/Integration/Filter/Storage/SessionFilterStorageTest.php
diimpp: My question was more in "do we need it?" dimension, because if we make this a standard, then tests codebase will double with tests like that `$this->assertTrue($container->has('sylius_grid.filter_storage'));`.

Shouldn't SF/DI tests cover this?

src/Component/Tests/Integration/Filter/Storage/SessionFilterStorageTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $container = $this->getContainer();
+
+        $this->assertTrue($container->has('sylius_grid.symfony.session.filter_storage'));
+        $this->assertInstanceOf(SessionFilterStorage::class, $container->get('sylius_grid.symfony.session.filter_storage'));
+
+        $this->assertTrue($container->has('sylius_grid.filter_storage'));
+        $this->assertInstanceOf(SessionFilterStorage::class, $container->get('sylius_grid.symfony.session.filter_storage'));
+
+        $this->assertTrue($container->has(FilterStorageInterface::class));
+        $this->assertInstanceOf(SessionFilterStorage::class, $container->get(FilterStorageInterface::class));
```
diimpp:
```suggestion
        $this->assertTrue($container->has('sylius.grid.filter_storage.session'));
        $this->assertInstanceOf(SessionFilterStorage::class, $container->get('sylius.grid.filter_storage.session'));

        $this->assertTrue($container->has('sylius.grid.filter_storage'));
        $this->assertInstanceOf(SessionFilterStorage::class, $container->get('sylius.grid.filter_storage.session'));

        $this->assertTrue($container->has(FilterStorageInterface::class));
        $this->assertInstanceOf(SessionFilterStorage::class, $container->get(FilterStorageInterface::class));
```

GSadee (approved):

PR #345, opened 2024-10-08
Update RegisterFiltersPass.php

ashceo:
In v1.13 filters which has implemented ```FilterInterface``` are autoregistered, but for this case:

```yaml
    sylius.grid_filter.enum:
        class: App\Component\Grid\Filter\EnumFilter
```

in compiler pass will be empty attribute from $container->registerForAutoconfiguration(FilterInterface::class) and exception is thrown on line #48 even the class has impletemented all methods from ConfigurableFilterInterface. In case when service ID is not an FQCN we should get service class.


GSadee (approved):

GSadee:
Thank you @dotdevio! 🎉

PR #346, opened 2024-10-09
Fix Doctrine optional requirement

loic425:
Maybe should we put this on 1.14 (next release), that will be safer
Edit: I've changed the base branch.

### DIFF
--- a/src/Bundle/Resources/config/services/integrations/doctrine/phpcr-odm.xml
+++ b/src/Bundle/Resources/config/services/integrations/doctrine/phpcr-odm.xml
@@ -16,9 +16,9 @@
         <defaults public="true" />
 
         <service id="sylius.grid_driver.doctrine.phpcrodm" class="Sylius\Bundle\GridBundle\Doctrine\PHPCRODM\Driver">
-            <argument type="service" id="doctrine_phpcr.odm.document_manager" />
+            <argument type="service" id="doctrine_phpcr.odm.document_manager" on-invalid="null" />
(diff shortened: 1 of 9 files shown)


src/Bundle/Resources/config/services/integrations/doctrine/phpcr-odm.xml
```diff
@@ -16,9 +16,9 @@
         <defaults public="true" />
 
         <service id="sylius.grid_driver.doctrine.phpcrodm" class="Sylius\Bundle\GridBundle\Doctrine\PHPCRODM\Driver">
-            <argument type="service" id="doctrine_phpcr.odm.document_manager" />
+            <argument type="service" id="doctrine_phpcr.odm.document_manager" on-invalid="ignore" />
```
loic425: The package could be available but the service could me missing if odm is not configured in Doctrine bundle.

src/Bundle/Resources/config/services/integrations/doctrine/phpcr-odm.xml
```diff
@@ -16,9 +16,9 @@
         <defaults public="true" />
 
         <service id="sylius.grid_driver.doctrine.phpcrodm" class="Sylius\Bundle\GridBundle\Doctrine\PHPCRODM\Driver">
-            <argument type="service" id="doctrine_phpcr.odm.document_manager" />
+            <argument type="service" id="doctrine_phpcr.odm.document_manager" on-invalid="null" />
```
loic425, line 19: the package could be available but not configured in doctrine bundle.

PR #348, opened 2024-10-21
Introduce AsFilter attribute to define filter type and form type

vasilvestre:
This PR introduces a new attribute, `AsFilter`, to provide a more concise and structured way of defining filter types and form types for grid filters.

Previously, filter types and form types were defined two methods inherited from `ConfigurableFilterInterface`.

By using the `AsFilter` attribute, developers can now explicitly declare the type and form type of a filter directly above the class definition. This approach improves code readability, reduces reliance on tag attributes, and enhances maintainability.

**Steps:**
- [x] Create the attribute and implement it's logic
- [x] Test
- [ ] Documentation (should be on Sylius stack now)

**Example usage:**

```php
use Sylius\Component\Grid\Filtering\FilterInterface;
use Sylius\Component\Grid\Metadata\AsFilter;

#[AsFilter(type: 'product_name', formType: ProductNameFilterType::class)]
final class ProductNameFilter implements FilterInterface
{
    public function apply(DataSourceInterface $dataSource, string $name, $data, array $options): void
    {
        if (empty($data['value'])) {
            return;
        }

        $queryBuilder = $dataSource->getQueryBuilder();
        $rootAlias = $queryBuilder->getRootAliases()[0];

        $queryBuilder
            ->andWhere(sprintf('%s.name LIKE :name', $rootAlias))
            ->setParameter('name', '%' . $data['value'] . '%');
    }
}
```

The `RegisterFiltersPass` compiler pass has been updated to check for the `AsFilter` attribute and ignore duplicate filter configuration.

Tests are provided in the PR, documentation too.

A new behaviour is that type is now optional as the FQCN is now used by default. Which means you can reference it directly in [PHP](https://github.com/Sylius/SyliusGridBundle/pull/348/files#diff-77aaeac60781f223afad001a87631ddf56d09651d5263d8b629f58c4d7354996R36) or [YAML](https://github.com/Sylius/SyliusGridBundle/pull/348/files#diff-96c3d347bc202eff3ed01bc089a7e1980297ccf35751e855bbc5049a1d5bd91fR111) .

**Future considerations:**

In a future PR, we can consider deprecating the `ConfigurableFilterInterface` as it becomes redundant by introducing the `AsFilter` attribute. This will further simplify the process of creating grid filters and improve consistency.

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
@@ -14,10 +14,12 @@ declare(strict_types=1);
 namespace Sylius\Bundle\GridBundle\DependencyInjection\Compiler;
 
 use InvalidArgumentException;
+use Sylius\Component\Grid\Attribute\AsFilter;
 use Sylius\Component\Grid\Filtering\FormTypeAwareFilterInterface;
 use Sylius\Component\Grid\Filtering\TypeAwareFilterInterface;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Reference;
 
 final class RegisterFiltersPass implements CompilerPassInterface
@@ -28,10 +30,21 @@ final class RegisterFiltersPass implements CompilerPassInterface
             return;
         }
 
-        $registry = $container->getDefinition('sylius.registry.grid_filter');
+        $filterRegistry = $container->getDefinition('sylius.registry.grid_filter');
         $formTypeRegistry = $container->getDefinition('sylius.form_registry.grid_filter');
 
-        foreach ($container->findTaggedServiceIds('sylius.grid_filter') as $id => $attributes) {
+        foreach ($container->findTaggedServiceIds(AsFilter::SERVICE_TAG) as $id => $attributes) {
+            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes);
+        }
+
+        foreach ($container->findTaggedServiceIds('sylius.legacy_grid_filter') as $id => $attributes) {
+            $definition = $container->findDefinition($id);
+
+            // Already configured with "sylius.grid_filter" tag.
+            if ($definition->hasTag(AsFilter::SERVICE_TAG)) {
+                continue;
+            }
+
             $type = null;
             $formType = null;
 
@@ -46,18 +59,29 @@ final class RegisterFiltersPass implements CompilerPassInterface
                 $formType = $class::getFormType();
             }
 
-            foreach ($attributes as $attribute) {
-                if (null === $type && null === ($attribute['type'] ?? null)) {
-                    throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "type" attributes or implements "%s".', TypeAwareFilterInterface::class));
-                }
+            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes, $type, $formType);
+        }
+    }
 
-                if (null === $formType && null === ($attribute['form_type'] ?? null)) {
-                    throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attributes or implements %s.', FormTypeAwareFilterInterface::class));
-                }
-
-                $registry->addMethodCall('register', [$type ?? $attribute['type'], new Reference($id)]);
-                $formTypeRegistry->addMethodCall('add', [$type ?? $attribute['type'], 'default', $formType ?? $attribute['form_type']]);
+    private function registerFilter(
+        Definition $filterRegistry,
+        Definition $formTypeRegistry,
+        string $id,
+        array $attributes,
+        ?string $type = null,
+        ?string $formType = null,
+    ): void {
+        foreach ($attributes as $attribute) {
+            if (null === $type && null === ($attribute['type'] ?? null)) {
+                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "type" attribute or implements "%s".', TypeAwareFilterInterface::class));
             }
+
+            if (null === $formType && null === ($attribute['form_type'] ?? null)) {
+                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attribute or implements "%s".', FormTypeAwareFilterInterface::class));
+            }
+
+            $filterRegistry->addMethodCall('register', [$type ?? $attribute['type'], new Reference($id)]);
+            $formTypeRegistry->addMethodCall('add', [$type ?? $attribute['type'], 'default', $formType ?? $attribute['form_type']]);
         }
     }
 }
--- a/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\DependencyInjection\Compiler;
 
-use InvalidArgumentException;
 use Sylius\Component\Grid\Attribute\AsFilter;
 use Sylius\Component\Grid\Filtering\FormTypeAwareFilterInterface;
 use Sylius\Component\Grid\Filtering\TypeAwareFilterInterface;
@@ -30,11 +29,11 @@ final class RegisterFiltersPass implements CompilerPassInterface
             return;
         }
 
-        $filterRegistry = $container->getDefinition('sylius.registry.grid_filter');
+        $registry = $container->getDefinition('sylius.registry.grid_filter');
         $formTypeRegistry = $container->getDefinition('sylius.form_registry.grid_filter');
 
         foreach ($container->findTaggedServiceIds(AsFilter::SERVICE_TAG) as $id => $attributes) {
-            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes);
+            $this->registerFilter($registry, $formTypeRegistry, $id, $attributes);
         }
 
         foreach ($container->findTaggedServiceIds('sylius.legacy_grid_filter') as $id => $attributes) {
@@ -59,7 +58,7 @@ final class RegisterFiltersPass implements CompilerPassInterface
                 $formType = $class::getFormType();
             }
 
-            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes, $type, $formType);
+            $this->registerFilter($registry, $formTypeRegistry, $id, $attributes, $type, $formType);
         }
     }
 
@@ -73,11 +72,11 @@ final class RegisterFiltersPass implements CompilerPassInterface
     ): void {
         foreach ($attributes as $attribute) {
             if (null === $type && null === ($attribute['type'] ?? null)) {
-                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "type" attribute or implements "%s".', TypeAwareFilterInterface::class));
+                throw new \InvalidArgumentException(sprintf('Tagged grid filters needs to have "type" attribute or implements "%s".', TypeAwareFilterInterface::class));
             }
 
             if (null === $formType && null === ($attribute['form_type'] ?? null)) {
-                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attribute or implements "%s".', FormTypeAwareFilterInterface::class));
+                throw new \InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attribute or implements "%s".', FormTypeAwareFilterInterface::class));
             }
 
             $filterRegistry->addMethodCall('register', [$type ?? $attribute['type'], new Reference($id)]);
--- a/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
@@ -29,26 +29,15 @@ final class RegisterFiltersPass implements CompilerPassInterface
             return;
         }
 
-        $registry = $container->getDefinition('sylius.registry.grid_filter');
+        $filterRegistry = $container->getDefinition('sylius.registry.grid_filter');
         $formTypeRegistry = $container->getDefinition('sylius.form_registry.grid_filter');
 
         foreach ($container->findTaggedServiceIds(AsFilter::SERVICE_TAG) as $id => $attributes) {
-            $this->registerFilter($registry, $formTypeRegistry, $id, $attributes);
-        }
-
-        foreach ($container->findTaggedServiceIds('sylius.legacy_grid_filter') as $id => $attributes) {
-            $definition = $container->findDefinition($id);
-
-            // Already configured with "sylius.grid_filter" tag.
-            if ($definition->hasTag(AsFilter::SERVICE_TAG)) {
-                continue;
-            }
+            $definition = $container->getDefinition($id);
+            $class = $definition->getClass();
 
             $type = null;
             $formType = null;
-
-            $definition = $container->getDefinition($id);
-            $class = $definition->getClass();
 
             if ($class !== null && is_a($class, TypeAwareFilterInterface::class, true)) {
                 $type = $class::getType();
@@ -58,7 +47,7 @@ final class RegisterFiltersPass implements CompilerPassInterface
                 $formType = $class::getFormType();
             }
 
-            $this->registerFilter($registry, $formTypeRegistry, $id, $attributes, $type, $formType);
+            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes, $type, $formType);
         }
     }
 
--- a/src/Bundle/DependencyInjection/SyliusGridExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusGridExtension.php
@@ -78,7 +78,7 @@ final class SyliusGridExtension extends Extension
         );
 
         $container->registerForAutoconfiguration(FilterInterface::class)
-            ->addTag('sylius.grid_filter')
+            ->addTag('sylius.legacy_grid_filter')
         ;
 
         $container->registerForAutoconfiguration(DataProviderInterface::class)
--- a/src/Bundle/DependencyInjection/SyliusGridExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusGridExtension.php
@@ -71,8 +71,8 @@ final class SyliusGridExtension extends Extension
                 }
 
                 $definition->addTag(AsFilter::SERVICE_TAG, [
-                'type' => $attribute->getType() ?? $reflector->getName(),
-                'form_type' => $attribute->getFormType(),
+                    'type' => $attribute->type ?? $reflector->getName(),
+                    'form_type' => $attribute->formType,
                 ]);
             },
         );
--- a/src/Bundle/DependencyInjection/SyliusGridExtension.php
+++ b/src/Bundle/DependencyInjection/SyliusGridExtension.php
@@ -16,9 +16,11 @@ namespace Sylius\Bundle\GridBundle\DependencyInjection;
 use Sylius\Bundle\CurrencyBundle\SyliusCurrencyBundle;
 use Sylius\Bundle\GridBundle\Grid\GridInterface;
 use Sylius\Bundle\GridBundle\SyliusGridBundle;
+use Sylius\Component\Grid\Attribute\AsFilter;
 use Sylius\Component\Grid\Data\DataProviderInterface;
 use Sylius\Component\Grid\Filtering\FilterInterface;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\Extension;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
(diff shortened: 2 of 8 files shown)


src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,6 +62,14 @@ public function load(array $configs, ContainerBuilder $container): void
             ->addTag('sylius.grid')
         ;
 
+        $container->registerAttributeForAutoconfiguration(AsFilter::class, static function (ChildDefinition $definition, AsFilter $attribute, \ReflectionClass $reflector) {
+            $tagAttributes = [
+                'type' => $attribute->type,
```
loic425: It could be cool to have the class name as the default type name.
```suggestion
                'type' => $attribute->type ?? $reflector->getName(),
```

vasilvestre: The FQCN is now the default type, I used it here https://github.com/Sylius/SyliusGridBundle/pull/348/files#diff-77aaeac60781f223afad001a87631ddf56d09651d5263d8b629f58c4d7354996R36

vasilvestre: The doc and tests uses the new FQCN behaviour

tests/Application/src/Filter/AttributeNationalityFilter.php
```diff
@@ -0,0 +1,35 @@
...
+namespace App\Filter;
+
+use App\Grid\Type\NationalityFilterType;
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Filter\EntityFilter;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+use Sylius\Component\Grid\Metadata\AsFilter;
+
+#[AsFilter(
+    formType: NationalityFilterType::class,
```
loic425, line 23: It's beautiful

tests/Application/src/Grid/Builder/AttributeNationalityFilter.php
```diff
@@ -0,0 +1,39 @@
...
+use App\Entity\Nationality;
+use App\Grid\Type\NationalityFilterType;
+use Sylius\Bundle\GridBundle\Builder\Filter\Filter;
+use Sylius\Bundle\GridBundle\Builder\Filter\FilterInterface;
+
+final class AttributeNationalityFilter
+{
+    public static function create(string $name, ?bool $multiple = null, ?array $fields = null): FilterInterface
+    {
+        $filter = Filter::create($name, $name);
```
loic425: I think we can directly call NationalityFilter::create method.

src/Component/Metadata/AsFilter.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsFilter
+{
+    /**
+     * @param string $formType The form type class name
```
loic425:
```suggestion
     * @param string $formType The form type class name to use for filter rendering
```

tests/Application/src/Grid/Builder/AttributeNationalityFilter.php
vasilvestre: The NationalityFilter is the example that's here to ensure using ConfigureableFilter still works, they are not linked together.
NationalityFilter actually use Filter::create() method.

vasilvestre: Improved with ::class const

loic425 (dismissed):

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
```diff
@@ -44,6 +44,10 @@ public function process(ContainerBuilder $container): void
             }
 
             foreach ($attributes as $attribute) {
+                if (isset($attribute['type'], $attribute['form_type'])) {
```
Roshyo: Isn't it?
```suggestion
                if (!isset($attribute['type'], $attribute['form_type'])) {
```

vasilvestre: No, I check the presence of a type and form_type attribute to avoid reconfiguring a filter that already has with Attribute. To keep a compatibility with the previous Sylius versions, we need to keep the autowire of all FilterInterface which leads to this issue.

vasilvestre: In the case of Attribute and FilterInterace, the $attributes will look like this :

```php
// Attribute then Interface data (unconfigured)
[
  'type' => 'foo',
  'form_type' => FormType::Class
],
[]
```

The first iteration can be skipped to the second part of the array but it means you register your filter twice which is never wanted AFAIK. (@loic425 confirmed that too)

I'm open to suggestion but the BC thing is annoying in this case.

Roshyo: Ha right, got it ;)

Roshyo (approved):

src/Component/Metadata/AsFilter.php
diimpp, line 1: About attribute class placement, should we follow sylius/resource example with `Metadata` or sylius-2.0 example with `Bundle/Attribute`?

https://github.com/Sylius/Sylius/blob/c9376c39aaa25a0818b62fbb6c86eedf94f367af/src/Sylius/Bundle/PromotionBundle/Attribute/AsPromotionCouponEligibilityChecker.php#L14
https://github.com/Sylius/Sylius/blob/c9376c39aaa25a0818b62fbb6c86eedf94f367af/src/Sylius/Bundle/ShippingBundle/Attribute/AsShippingCalculator.php#L14

@loic425 what do you think?

Personally I'm in for `src/Bundle/Attribute/AsFilter` since this is a fancy service registration and not part of complex metadata configuration as it goes for resource.

Plus it's mostly bundle related and not to component.

src/Component/Metadata/AsFilter.php
```diff
@@ -0,0 +1,28 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Metadata;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsFilter
+{
+    /**
+     * @param string $formType The form type class name to use for filter rendering
```
diimpp:
```suggestion
     * @param class-string $formType The form type class name to use for filter rendering
```

src/Component/Metadata/AsFilter.php
```diff
@@ -0,0 +1,28 @@
...
+     * @param string|null $type The filter type (defaults to FQCN of the class)
```
diimpp: `@param` which duplicates type hint without additional clarification for type like `class-string` shouldn't be added, in fact it will be removed with next CS fixes run. (Unless grid CS config outdated).

```suggestion
```

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,6 +62,17 @@ public function load(array $configs, ContainerBuilder $container): void
             ->addTag('sylius.grid')
         ;
 
+        $container->registerAttributeForAutoconfiguration(AsFilter::class, function (ChildDefinition $definition, AsFilter $attribute, \Reflector $reflector): void {
+            if (!$reflector instanceof \ReflectionClass) {
```
diimpp: Do you still need to use reflection if you've specified target on the attribute?

image

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,6 +62,17 @@ public function load(array $configs, ContainerBuilder $container): void
             ->addTag('sylius.grid')
         ;
 
+        $container->registerAttributeForAutoconfiguration(AsFilter::class, function (ChildDefinition $definition, AsFilter $attribute, \Reflector $reflector): void {
```
diimpp: Let's follow established syntax at sylius-2.0

https://github.com/Sylius/Sylius/blob/c9376c39aaa25a0818b62fbb6c86eedf94f367af/src/Sylius/Bundle/PromotionBundle/DependencyInjection/SyliusPromotionExtension.php#L52-L62

src/Component/Metadata/AsFilter.php
```diff
@@ -0,0 +1,28 @@
...
+     */
+    public function __construct(
+        public string $formType,
+        public ?string $type = null,
```
diimpp:
```suggestion
        private string $formType,
        private ?string $type = null,
```

diimpp:
```php
    public function getFormType(): string
    {
        return $this->formType;
    }

    public function getType(): ?string
    {
        return $this->type;
    }
```

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
```diff
@@ -44,6 +44,10 @@ public function process(ContainerBuilder $container): void
...
+                    return;
```
diimpp: Shouldn't we continue with the rest of attributes, even if current one is already registered?
```suggestion
                    continue;
```

diimpp (changes requested):

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
vasilvestre: Explanation here :

https://github.com/Sylius/SyliusGridBundle/pull/348#discussion_r1810457553

src/Component/Metadata/AsFilter.php
vasilvestre: I liked the documentation but let's remove it

src/Bundle/DependencyInjection/SyliusGridExtension.php
vasilvestre: I did this for static analysis. \Reflector->getName() does not exist, that's why I ensure we use the \ReflectionClass. Any better idea ?

Call to an undefined method Reflector::getName().   

I don't like to add phpdoc even when we're "sure" the attribute targets a class, if this changes then the phpdoc will be false and misleading.

src/Component/Metadata/AsFilter.php
loic425: That's true for Sylius monorepository, but when you look at Symfony components, there are in components.
https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Console/Attribute/AsCommand.php

src/Bundle/DependencyInjection/SyliusGridExtension.php
loic425: yes, @vasilvestre specified the attribute to be used only on class.

src/Component/Metadata/AsFilter.php
loic425: public is perfect for attributes
https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/Console/Attribute/AsCommand.php

In resource metadata, it's more complicated cause we use the object as attribute and as object we manipulate.

src/Component/Metadata/AsFilter.php
vasilvestre: So I should move it to Attribute namespace ?

loic425:
@vasilvestre you have two commits with "-" as commit message
https://github.com/Sylius/SyliusGridBundle/pull/348/commits

src/Component/Metadata/AsFilter.php
loic425: yes, attribute namespace will be better.
I think I will move attributes on resource at some point.

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
vasilvestre: Apparently this now works, tests are ok

src/Component/Metadata/AsFilter.php
vasilvestre: I also like the public approach more but we can't use readonly on them because we set these field while processing them

loic425: hum... ok, but public is ok for me.
So remove the readonly on the other PR too to make this consistant.

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,6 +62,21 @@ public function load(array $configs, ContainerBuilder $container): void
...
+            static function (ChildDefinition $definition, AsFilter $attribute, \Reflector $reflector): void {
+                // Helps to avoid issues with psalm
+                if (!$reflector instanceof \ReflectionClass) {
+                    return;
+                }
+
+                $definition->addTag(AsFilter::SERVICE_TAG, [
+                'type' => $attribute->getType() ?? $reflector->getName(),
+                'form_type' => $attribute->getFormType(),
+                ]);
```
diimpp:
```suggestion
                $definition->addTag(AsFilter::SERVICE_TAG, [
                    'type' => $attribute->getType() ?? $reflector->getName(),
                    'form_type' => $attribute->getFormType(),
                ]);
```

src/Component/Attribute/AsFilter.php
```diff
@@ -0,0 +1,39 @@
...
+final class AsFilter
+{
+    public const SERVICE_TAG = 'sylius.grid.filter';
+
+    /**
+     * @param class-string $formType The form type class name to use for filter rendering
+     */
+    public function __construct(
+        private readonly string $formType,
+        private readonly ?string $type = null,
```
diimpp, line 26:
```suggestion
        private string $formType,
        private ?string $type = null,
```

src/Component/Metadata/AsFilter.php
vasilvestre: So I changed to public not readonly as both loic and Łukasz agreed, are you ok with it @diimpp ?

diimpp (approved):

tests/Application/config/sylius/grids/book.php
```diff
@@ -31,6 +32,11 @@
             null,
             ['author.nationality'],
         ))
+        ->addFilter(AttributeNationalityFilter::create(
+            AttributeNationalityFilter::class,
```
lchrusciel, line 36: it seems to be incorrect, or at least misleading

tests/Application/config/sylius/grids/book.php
```diff
@@ -31,6 +32,11 @@
             null,
             ['author.nationality'],
         ))
+        ->addFilter(AttributeNationalityFilter::create(
```
lchrusciel, line 35: How is it tested? Does it override nationality filter and is covered by this test: https://github.com/Sylius/SyliusGridBundle/blob/f500a89f5ea35fd675012c30e55d0def0a751cc3/src/Bundle/Tests/Functional/GridUiTest.php#L235-L245 or is not covered by test and we don't know if it works?

tests/Application/config/sylius/grids/book.php
vasilvestre: Why ? We can use FQCN as name and also call the create method of the filter.

tests/Application/config/sylius/grids/book.php
vasilvestre: I think it's covered in https://github.com/Sylius/SyliusGridBundle/blob/f500a89f5ea35fd675012c30e55d0def0a751cc3/tests/Application/src/Filter/AttributeNationalityFilter.php and https://github.com/Sylius/SyliusGridBundle/blob/f500a89f5ea35fd675012c30e55d0def0a751cc3/tests/Application/src/Grid/Builder/AttributeNationalityFilter.php

tests/Application/config/sylius/grids/book.php
loic425: Yes, this is a bit strange... but I understand why.
The FQCN is the type/name of the filter.

loic425: @vasilvestre Could you remember me why you don't use the Filter::create method instead?

vasilvestre: I did line 35

src/Component/Attribute/AsFilter.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsFilter
+{
+    public const SERVICE_TAG = 'sylius.grid.filter';
```
loic425:
```suggestion
    public const SERVICE_TAG = 'sylius.grid_filter';
```

loic425 (changes requested):
Does not change it yourself, I have tests in progress.

loic425 (approved):

tests/Application/config/sylius/grids/book.php
loic425: I added tests on the RegisterFiltersPass itself.

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,8 +62,23 @@ public function load(array $configs, ContainerBuilder $container): void
...
+                $definition->addTag(AsFilter::SERVICE_TAG, [
+                    'type' => $attribute->type ?? $reflector->getName(),
+                    'form_type' => $attribute->formType,
+                ]);
+            },
+        );
+
         $container->registerForAutoconfiguration(FilterInterface::class)
-            ->addTag('sylius.grid_filter')
+            ->addTag('sylius.legacy_grid_filter')
```
loic425: We now tagged the Attribute instead, but we need the ones that implements TypeAwareFilterInterface and FormTypeAwareFilterInterface which contains static methods for the type and the form type.

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
```diff
@@ -28,10 +30,21 @@ public function process(ContainerBuilder $container): void
             return;
         }
 
-        $registry = $container->getDefinition('sylius.registry.grid_filter');
+        $filterRegistry = $container->getDefinition('sylius.registry.grid_filter');
```
diimpp, line 32: If `registry` is renamed to `filterRegistry`, then `formTypeRegistry` should be become `filterFormTypeRegister`, hence I propose to remove this change.

```suggestion
        $registry = $container->getDefinition('sylius.registry.grid_filter');
```

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
```diff
@@ -46,18 +59,29 @@ public function process(ContainerBuilder $container): void
...
+    ): void {
+        foreach ($attributes as $attribute) {
+            if (null === $type && null === ($attribute['type'] ?? null)) {
+                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "type" attribute or implements "%s".', TypeAwareFilterInterface::class));
+            }
 
-                $registry->addMethodCall('register', [$type ?? $attribute['type'], new Reference($id)]);
-                $formTypeRegistry->addMethodCall('add', [$type ?? $attribute['type'], 'default', $formType ?? $attribute['form_type']]);
+            if (null === $formType && null === ($attribute['form_type'] ?? null)) {
+                throw new InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attribute or implements "%s".', FormTypeAwareFilterInterface::class));
```
diimpp: According to SF CS, which is basis for sylius CS, global classes should be used directly and not added to `use`.

```suggestion
                throw new \InvalidArgumentException(sprintf('Tagged grid filters needs to have "form_type" attribute or implements "%s".', FormTypeAwareFilterInterface::class));
```

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
```diff
@@ -14,10 +14,12 @@
 namespace Sylius\Bundle\GridBundle\DependencyInjection\Compiler;
 
 use InvalidArgumentException;
```
diimpp:
```suggestion
```

tests/Application/src/Filter/AttributeNationalityFilter.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace App\Filter;
+
+use App\Grid\Type\NationalityFilterType;
```
diimpp, line 16: Please move form type to usual SF placement.

```suggestion
use App\Form\Type\Grid\NationalityFilterType;
```

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -60,8 +62,23 @@ public function load(array $configs, ContainerBuilder $container): void
...
+                }
+
+                $definition->addTag(AsFilter::SERVICE_TAG, [
+                    'type' => $attribute->type ?? $reflector->getName(),
+                    'form_type' => $attribute->formType,
+                ]);
+            },
+        );
+
         $container->registerForAutoconfiguration(FilterInterface::class)
```
loic425: we should use this interface instead, that was a mistake to use it cause the FilterInterface itself is not enough to auconfigure it.

```suggestion
        $container->registerForAutoconfiguration(ConfigurableFilterInterface::class)
```

tests/Application/src/Filter/AttributeNationalityFilter.php
vasilvestre: This is not the type of the already existing filter. Last change to this filter is from 2022, I'm not sure this is super useful to make this change in this PR.

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
vasilvestre: I like the change

src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
vasilvestre: Shouldn't this be enforced by our CS rules ?

loic425:
@diimpp & @vasilvestre 
Previously it was the FilterInterface which was autoconfigured, but it was a mistake and was commented by Lukasz during review when the ConfigurableFilterInterface feature was implemented. With only this interface, you wasn't able to have the correct tag with the needed attributes.
So I simply autoconfigure the ConfigurableFilterInterface but also detect if it is already configured by the AsFilter first.

I know this is not simple at all... but I think it's safe like this and I've added some tests on the RegisterFilterPassTest file.

The original PR where the Grid filter autoconfiguration feature was introduced
https://github.com/Sylius/SyliusGridBundle/pull/276

Edit:
I've tried on a customer project, and indeed, if we autoconfigure the FilterInterface it does not work. we have to unconfigure it to define the form type and type manually.
```php
$services->set(ClientFilter::class)
        ->autoconfigure(false)
        ->tag(
            name: 'sylius.grid_filter',
            attributes: [
                'form_type' => $clientFilterType,
                'type' => ClientFilter::class,
            ],
        )
    ;
```

GSadee (approved):

GSadee:
Thank you @vasilvestre and @loic425! 🎉

PR #349, opened 2024-10-22
Introduce AsGrid attribute for streamlined grid configuration

vasilvestre:
To simplify the configuration of grids and promote code clarity, this PR introduces a new attribute, `AsGrid`, which allows developers to define the grid's name and resource class directly within their grid classes, eliminating the need for YAML configuration files.

Previously, these properties were defined at usage which could cause code duplication.

By using the `AsGrid` attribute, developers can now explicitly declare the name and resource class of a grid directly above the class definition. This approach improves code readability, centralizes configuration, and enhances maintainability.

**Example usage:**

```php
use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
use Sylius\Component\Grid\Metadata\AsGrid;

#[AsGrid(
    resourceClass: Product::class, 
    name: 'app_admin_product',
)]
final class ProductGrid implements AbstractGrid
{
    // ...
}
```

```php
use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
use Sylius\Component\Grid\Metadata\AsGrid;

#[AsGrid]
final class ProductGrid implements AbstractGrid
{
    // ...
}
```

This PR includes comprehensive tests and updated documentation to guide developers on using the new attribute.

**Future considerations:**

In a future PR, we can consider deprecating the YAML-based grid configuration in favor of the attribute-based approach to further streamline the grid creation process and improve consistency.

### DIFF
--- a/src/Bundle/Grid/AbstractGrid.php
+++ b/src/Bundle/Grid/AbstractGrid.php
@@ -22,7 +22,7 @@ abstract class AbstractGrid implements GridInterface
     public static function getName(): string
     {
         if ($attribute = self::getAttributes()) {
-            return $attribute[0]->newInstance()->getName();
+            return $attribute[0]?->newInstance()->name ?? static::class;
         }
 
         return static::class;
@@ -31,7 +31,7 @@ abstract class AbstractGrid implements GridInterface
     public function getResourceClass(): string
     {
         if ($attribute = self::getAttributes()) {
-            return $attribute[0]->newInstance()->getResourceClass();
+            return $attribute[0]->newInstance()->resourceClass;
         }
 
         throw new \LogicException(sprintf(
--- a/src/Bundle/Grid/AbstractGrid.php
+++ b/src/Bundle/Grid/AbstractGrid.php
@@ -15,9 +15,32 @@ namespace Sylius\Bundle\GridBundle\Grid;
 
 use Sylius\Bundle\GridBundle\Builder\GridBuilder;
 use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
+use Sylius\Component\Grid\Attribute\AsGrid;
 
 abstract class AbstractGrid implements GridInterface
 {
+    public static function getName(): string
+    {
+        if ($attribute = self::getAttributes()) {
+            return $attribute[0]->newInstance()->getName();
+        }
+
+        return static::class;
+    }
+
+    public function getResourceClass(): string
+    {
+        if ($attribute = self::getAttributes()) {
+            return $attribute[0]->newInstance()->getResourceClass();
+        }
+
+        throw new \LogicException(sprintf(
+            'You have to implement %s method or use %s attribute.',
+            __METHOD__,
+            AsGrid::class,
+        ));
+    }
+
     public function toArray(): array
     {
         $gridBuilder = $this->createGridBuilder();
@@ -29,10 +52,18 @@ abstract class AbstractGrid implements GridInterface
 
     private function createGridBuilder(): GridBuilderInterface
     {
-        if ($this instanceof ResourceAwareGridInterface) {
+        if ($this instanceof ResourceAwareGridInterface || $this::getAttributes() !== []) {
             return GridBuilder::create($this::getName(), $this->getResourceClass());
         }
 
         return GridBuilder::create($this::getName());
+    }
+
+    /**
+     * @return \ReflectionAttribute<AsGrid>[]
+     */
+    private static function getAttributes(): array
+    {
+        return (new \ReflectionClass(static::class))->getAttributes(AsGrid::class);
     }
 }
--- a/src/Bundle/Grid/AbstractGrid.php
+++ b/src/Bundle/Grid/AbstractGrid.php
@@ -21,24 +21,7 @@ abstract class AbstractGrid implements GridInterface
 {
     public static function getName(): string
     {
-        if ($attribute = self::getAttributes()) {
-            return $attribute[0]?->newInstance()->name ?? static::class;
-        }
-
-        return static::class;
-    }
-
-    public function getResourceClass(): string
-    {
-        if ($attribute = self::getAttributes()) {
-            return $attribute[0]->newInstance()->resourceClass;
-        }
-
-        throw new \LogicException(sprintf(
-            'You have to implement %s method or use %s attribute.',
-            __METHOD__,
-            AsGrid::class,
-        ));
+        return self::getAsGridAttribute()?->name ?? static::class;
     }
 
     public function toArray(): array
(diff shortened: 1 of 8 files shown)


src/Bundle/Grid/AbstractGrid.php
```diff
@@ -15,9 +15,32 @@
...
 
 abstract class AbstractGrid implements GridInterface
 {
+    public static function getName(): string
+    {
+        if ($attribute = (new \ReflectionClass(static::class))->getAttributes(AsGrid::class)) {
+            return $attribute[0]->newInstance()->name;
+        }
+
+        return static::getName();
```
loic425: self::class or static::class not sure

src/Component/Attribute/AsGrid.php
```diff
@@ -0,0 +1,34 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsGrid
+{
+    public function __construct(
+        private readonly string $name,
```
loic425: I think we should pass it as public. readonly is enough to tackle this.

loic425: @diimpp public is perfect. It's the modern way. We don't need getters for this kind of things.
Moreover, Symfony tackle them like that too.

src/Bundle/Grid/AbstractGrid.php
```diff
@@ -15,9 +15,32 @@
...
+        return static::class;
+    }
+
+    public function getResourceClass(): string
+    {
+        if ($attribute = self::getAttributes()) {
+            return $attribute[0]->newInstance()->getResourceClass();
+        }
+
+        throw new \LogicException(sprintf(
```
loic425: we should add an exception namespace on Grid. But this need to be done in another PR.

src/Component/Attribute/AsGrid.php
lchrusciel: 👍 for Loic perspective. 

Also, I would investigate if we can even make name optional, as we can already work with FQCN for grids. Name is still required for Sylius as a framework, but on end app, FQCN should be enough

vasilvestre: Should I change it back so ?

loic425: Yep and indeed name can be optional and be the FQCN by default as you did for the filter attribute

vasilvestre: Done, tell me what you think

src/Bundle/Tests/DependencyInjection/GridBuilderConfigurationTest.php
```diff
@@ -903,6 +904,37 @@ public function it_builds_extended_grids_with_grids_as_service(): void
...
+    /**
+     * @test
+     */
+    public function it_builds_grid_with_a_grid_attribute(): void
+    {
+        $grid = new AttributeFooGrid();
+
+        $this->load([
+            'grids' => [
+                'app_foo_with_attribute' => $grid->toArray(),
```
diimpp: Should test cover case where name is not specified and FQCN is used instead?

vasilvestre: Smart ! PR updated

src/Component/Tests/Dummy/AttributeFooGrid.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Tests\Dummy;
+
+use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
+use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
+use Sylius\Component\Grid\Attribute\AsGrid;
+
+#[AsGrid(resourceClass: Foo::class)]
+final class AttributeFooGrid extends AbstractGrid
```
lchrusciel: Is it required to extend `AbstractGrid` to take advantage of `AsGrid` attribute?

src/Bundle/Grid/AbstractGrid.php
```diff
@@ -15,9 +15,34 @@
...
+        if ($attribute = self::getAttributes()) {
+            return $attribute[0]->newInstance()->resourceClass;
+        }
+
+        throw new \LogicException(sprintf(
+            'You have to implement %s method or use %s attribute.',
+            __METHOD__,
+            AsGrid::class,
+        ));
+    }
```
lchrusciel: Is this code needed to make `AsGrid` useful?

vasilvestre: Absolutely yes

src/Component/Tests/Dummy/AttributeFooGrid.php
vasilvestre: Nice catch, I will take a loot. AFAIK an interface should be enough

vasilvestre: So, the AbstractGrid fetches the name in attributes or via code.
If we use GridInterface, getName() and toArray() must be reimplemented.

loic425 (dismissed):

src/Bundle/Grid/AbstractGrid.php
Prometee: Can we use string parameter instead of the class? It can be useful for plugins.

loic425: Hum I'm wondering sth... This getResourceClass is on ResourceAwareGridInterface.

src/Component/Attribute/AsGrid.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class AsGrid
+{
+    public function __construct(
+        public readonly string $resourceClass,
```
loic425: It should be optional, cause grid can be unrelated to resource classes.

vasilvestre: Should we add test to configure a non resource related ?

src/Bundle/Grid/AbstractGrid.php
vasilvestre: I'm not sure what you are meaning, which line ?

stlgaits (approved):

src/Component/Tests/Dummy/AttributeFooGrid.php
loic425: Yes, the AbstractGrid allow to read the grid name from the AsGrid attribute.
The `toArray` method from AbstractGrid call the grid builder `build` method and call its `toArray` method.

diimpp (approved):

GSadee (approved):

GSadee:
Thank you @vasilvestre! 🎉

Issue #303, opened 2023-05-11
Grid sorting does not work without the sorting key being configured


vvasiloi:
If a grid has fields with sorting enabled, but without a `sorting` key configured, then the sorting won't work.
A prior discussion on this issue: https://github.com/Sylius/Sylius/issues/14131

PR #354, opened 2024-11-27
fix grid sorting with sorting key not defined and refactor grid definition creation

Florian-Merle:
This PR solves https://github.com/Sylius/SyliusGridBundle/issues/303 and refactors how a `Grid` is constructed. 

The `Grid` definition creation process is kinda duplicated in `ArrayGridProvider` and `ServiceGridProvider`.

I introduce a `GridFactory` to transform a grid configuration array into a `Grid`.

This service dispatch an event so the grid configuration array can be manipulated. Three listener have been added to reproduce what was done before and fix the sorting when the sorting key is not defined.
* `GridConfigurationExtenderListener`
* `GridConfigurationRemovalsListener`
* `GridConfigurationSortingListener`

PR #367, opened 2025-03-03
fix grid sorting with sorting key not defined and refactor grid definition creation

Florian-Merle:
This PR fixes issue #303  without refactoring the code as previously done in the PR #354

While refactoring would be nice, it seems complex without introducing BC breaks. That's why, this PR focuses on fixing the bug only


GSadee (approved):

GSadee:
Thank you @Florian-Merle! 🎉

PR #370, opened 2025-04-11
Configure template via the AsFilter attribute

loic425:
```php
#[AsFilter(
    formType: SourceFilterType::class,
    // This template configuration is the new feature implemented on that PR
    template: '@SyliusBootstrapAdminUi/shared/grid/filter/select.html.twig',
)]
final class SourceFilter implements FilterInterface
{
   // ...
}
```

```php
// config/packages/sylius_grid.php
return static function (ContainerConfigurator $containerConfigurator): void {
    $containerConfigurator->extension('sylius_grid', [
        'templates' => [
            'filter' => [
                // It allows to remove that configuration
                // SourceFilter::class => '@SyliusBootstrapAdminUi/shared/grid/filter/select.html.twig',
            ],
        ],
    ]);
};
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,6 +40,7 @@
         "symfony/http-kernel": "^5.4 || ^6.4 || ^7.0",
         "symfony/options-resolver": "^5.4 || ^6.4 || ^7.0",
         "symfony/property-access": "^5.4 || ^6.4 || ^7.0",
+        "symfony/validator": "^5.4 || ^6.4 || ^7.0",
         "webmozart/assert": "^1.9"
     },
     "replace": {
--- a/tests/Application/config/sylius/grids/book.php
+++ b/tests/Application/config/sylius/grids/book.php
@@ -34,7 +34,7 @@ return static function (GridConfig $grid) {
             ['author.nationality'],
         ))
         ->addFilter(AttributeNationalityFilter::create(
-            AttributeNationalityFilter::class,
+            'attribute_nationality',
             null,
             ['author.nationality'],
         ))
--- a/tests/Application/templates/grid/filter/string.html.twig
+++ b/tests/Application/templates/grid/filter/string.html.twig
@@ -1,0 +1,4 @@
+{% if form.type is defined %}
+    {{ form_row(form.type, {'label': filter.label}) }}
+{% endif %}
+    {{ form_row(form.value, {attr: {placeholder: form.value.vars.label}, 'label': form.type is defined ? form.value.vars.label : filter.label}) }}
--- a/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/RegisterFiltersPass.php
@@ -47,11 +47,12 @@ final class RegisterFiltersPass implements CompilerPassInterface
                 $formType = $class::getFormType();
             }
 
-            $this->registerFilter($filterRegistry, $formTypeRegistry, $id, $attributes, $type, $formType);
+            $this->registerFilter($container, $filterRegistry, $formTypeRegistry, $id, $attributes, $type, $formType);
         }
     }
 
     private function registerFilter(
+        ContainerBuilder $container,
         Definition $filterRegistry,
         Definition $formTypeRegistry,
         string $id,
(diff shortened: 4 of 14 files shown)


src/Bundle/Renderer/TwigGridRenderer.php
```diff
@@ -113,14 +113,14 @@ public function renderAction(GridViewInterface $gridView, Action $action, $data
     public function renderFilter(GridViewInterface $gridView, Filter $filter)
     {
         $template = $this->getFilterTemplate($filter);
+        $filterName = str_replace('\\', '', $filter->getName());
 
         $form = $this->formFactory->createNamed('criteria', FormType::class, [], [
-            'allow_extra_fields' => true,
```
loic425, line 118: I don't understand why this option was here.
In the test application, it breaks with this following error (but it works on all of our projects and I don't understand why).
Capture d’écran du 2025-04-11 11-59-11

src/Bundle/Renderer/TwigGridRenderer.php
```diff
@@ -113,14 +113,14 @@ public function renderAction(GridViewInterface $gridView, Action $action, $data
     public function renderFilter(GridViewInterface $gridView, Filter $filter)
     {
         $template = $this->getFilterTemplate($filter);
+        $filterName = str_replace('\\', '', $filter->getName());
```
loic425: When using the FQCN on filter name, it breaks the form name.
image

composer.json
```diff
@@ -40,6 +40,7 @@
         "symfony/http-kernel": "^5.4 || ^6.4 || ^7.0",
         "symfony/options-resolver": "^5.4 || ^6.4 || ^7.0",
         "symfony/property-access": "^5.4 || ^6.4 || ^7.0",
+        "symfony/validator": "^5.4 || ^6.4 || ^7.0",
```
loic425, line 43: needed to use allow_extra_fields option passed to the TwigGridRenderer
https://github.com/Sylius/SyliusGridBundle/blob/1.14/src/Bundle/Renderer/TwigGridRenderer.php#L118

So this is a bug fix.

src/Bundle/Renderer/TwigGridRenderer.php
vasilvestre: It seems a bit dirty sadly

src/Bundle/Renderer/TwigGridRenderer.php
```diff
@@ -113,14 +113,15 @@ public function renderAction(GridViewInterface $gridView, Action $action, $data
...
```
diimpp: Will this form name be visible anywhere, like SF debugger?

Should it be formatted to "some_name" or "someName" instead of current "SomeName"?

Like,
```suggestion
        $filterName = strtolower(str_replace('\\', '_', $filter->getName()));
```
or is there no way to see if it's FQCN instead of regular string?

diimpp (approved):

src/Bundle/Renderer/TwigGridRenderer.php
loic425: You're right, I think the name will be used for the subname of the criteria. `?criteria[form_name]=value`

tests/Application/config/sylius/grids/book.php
```diff
@@ -34,7 +34,7 @@
             ['author.nationality'],
         ))
         ->addFilter(AttributeNationalityFilter::create(
-            AttributeNationalityFilter::class,
+            'attribute_nationality',
```
loic425, line 37: I've removed FQCN as key to postpone that in another PR.

src/Bundle/Renderer/TwigGridRenderer.php
loic425: I've removed the FQCN for the attribute key support for now, we'll see that in another PR.

Wojdylak (approved):

tests/Application/templates/grid/filter/string.html.twig
```diff
@@ -0,0 +1,4 @@
+{% if form.type is defined %}
+    {{ form_row(form.type, {'label': filter.label}) }}
+{% endif %}
+    {{ form_row(form.value, {attr: {placeholder: form.value.vars.label}, 'label': form.type is defined ? form.value.vars.label : filter.label}) }}
```
GSadee, line 4:
```suggestion
{{ form_row(form.value, {attr: {placeholder: form.value.vars.label}, 'label': form.type is defined ? form.value.vars.label : filter.label}) }}
```

GSadee (approved):

GSadee:
Thank you @loic425! 🎉

PR #372, opened 2025-04-21
Fix grid sorting bc break

### LINKED ISSUE #369: BC break related to sorting in v1.14.0-ALPHA.1
BC break in https://github.com/Sylius/SyliusGridBundle/pull/367/files

This configuration does not work anymore because sortable is always set to true in https://github.com/Sylius/SyliusGridBundle/pull/367/files#diff-28dd6bf4f3e5c7489778da4e9684b4f522616b8eb21b276b27af8882e4fc4479R29

```yaml
sylius_grid:
    grids:
        tag:
            sorting:
                name: 'asc'
            fields:
                name:
                    sortable: 'translation.name'
```

Florian-Merle:
This fixes https://github.com/Sylius/SyliusGridBundle/issues/369

### DIFF
--- a/src/Component/Configuration/GridConfigurationSortingHandler.php
+++ b/src/Component/Configuration/GridConfigurationSortingHandler.php
@@ -26,6 +26,10 @@ final class GridConfigurationSortingHandler implements GridConfigurationSortingH
         foreach ($gridConfiguration['sorting'] as $sorting => $order) {
             Assert::keyExists($gridConfiguration['fields'] ?? [], $sorting);
 
+            if (isset($gridConfiguration['fields'][$sorting]['sortable'])) {
+                continue;
+            }
+
             $gridConfiguration['fields'][$sorting]['sortable'] = true;
         }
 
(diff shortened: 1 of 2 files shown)


src/Component/Configuration/GridConfigurationSortingHandler.php
```diff
@@ -26,6 +26,10 @@ public function handle(array $gridConfiguration): array
         foreach ($gridConfiguration['sorting'] as $sorting => $order) {
             Assert::keyExists($gridConfiguration['fields'] ?? [], $sorting);
 
+            if (isset($gridConfiguration['fields'][$sorting]['sortable'])) {
+                continue;
+            }
+
             $gridConfiguration['fields'][$sorting]['sortable'] = true;
         }
```
Florian-Merle, line 34: I'm wondering if using `isset` here is the right solution.

Looking at the tests, I had to update the line for the author because its sortable value was originally being changed from `false` to `true`. However, with these changes, that no longer happens.

Maybe the value should be set to `true` if it's either not set or currently `false`.

loic425: maybe should we throw an error that indicates the field is not sortable.

loic425:
@Florian-Merle Is this one ok for you?

loic425 (approved):

loic425:
@Florian-Merle I've tried the change on that grid and that fixes the issue.
https://github.com/Sylius/SyliusResourceBundle/pull/1019

diimpp (approved):

NoResponseMate (approved):

PR #373, opened 2025-04-23
fix: [make:grid Command] - make Doctrine ORM requirement optional

stlgaits:
This PR aims to prevent an exception being thrown when installing the SyliusGridBundle on an existing project without Doctrine ORM.

image

### DIFF
--- a/src/Bundle/Tests/Functional/Maker/MakeGridTest.php
+++ b/src/Bundle/Tests/Functional/Maker/MakeGridTest.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\GridBundle\Tests\Functional\Maker;
 
+use App\BoardGameBlog\Infrastructure\Sylius\Resource\BoardGameResource;
 use App\Entity\AdminUser;
 use App\Entity\Book;
 use App\Entity\Price;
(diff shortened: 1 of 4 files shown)


loic425 (approved):

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -43,6 +46,19 @@ public function it_can_create_grids(): void
         $this->assertSame(self::getPriceGridExpectedContent(), \file_get_contents(self::tempFile(self::PRICE_GRID_PATH)));
     }
 
+    /** @test */
+    public function it_can_create_grids_with_a_simple_resource(): void
```
diimpp: I think we don't have "simple resource" term, which will referrer to doctrine-less resource.

How about,
```patch
---    public function it_can_create_grids_with_a_doctrine_entity(): void
---    public function it_can_create_grids_with_a_simple_resource(): void
+++    public function it_can_create_grids: void
+++    public function it_can_create_grids_without_doctrine(): void
```

stlgaits: Changed it :)

diimpp (approved):

GSadee (approved):

GSadee:
Thank you @stlgaits! 🎉

PR #374, opened 2025-04-23
Improve make grid command to be run with Doctrine Entities or any other Object resource class

stlgaits:
The current implementation of the `make:grid` command assumes a Doctrine entity as the only valid input.
This PR introduces enhanced flexibility to the command by allowing developers to generate grids based on resource classes beyond standard Doctrine entities.

With this change, you can use any PHP class—such as a Sylius Resource implementation—as the basis for your grid configuration. This makes the command more versatile and better suited to projects using custom resource patterns or non-standard data sources.
Example usage:

```shell
symfony console make:grid 'App\BoardGameBlog\Infrastructure\\Sylius\Resource\BoardGameResource'
```

This command will now correctly scaffold a grid using the specified class, even if it is not a traditional Doctrine entity.

Based on #373

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 /composer.lock
 
 /.phpunit.result.cache
-src/Bundle/test/tmp
+tests/Application/tmp
 
 src/Bundle/test/vendor/
 src/Bundle/test/composer.lock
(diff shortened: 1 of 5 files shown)


diimpp (commented):
Command description needs to be updated as well.
```
    public static function getCommandDescription(): string
    {
        return 'Creates a Grid for a Doctrine entity class';
    }
```

@stlgaits is this PR replaces https://github.com/Sylius/SyliusGridBundle/pull/373 ?

diimpp:
> @stlgaits is this PR replaces #373 ?

I see now, 373 is a bugifx for 1.13 and 374 feature for 1.14

stlgaits:
@diimpp I changed the description to :  
```php

public static function getCommandDescription(): string
    {
        return 'Creates a Sylius Grid configuration for a given resource class or Doctrine entity.';
    }
```

WDYT?

diimpp (approved):

NoResponseMate (approved):

PR #375, opened 2025-05-06
Fix Twig & Doctrine integration only when available

loic425:
Twig is an optional requirement.
image

Based on #373 
Will replace #346 

The code is inspired on what I have done in SyliusResourceBundle (https://github.com/Sylius/SyliusResourceBundle/blob/1.13/src/Bundle/DependencyInjection/SyliusResourceExtension.php#L181).

Cause we handle the grid drivers differently, I think it's better to implement this into a new minor version (1.14).


src/Bundle/DependencyInjection/Configuration.php
```diff
@@ -40,7 +40,7 @@ private function addDriversSection(ArrayNodeDefinition $node): void
         $node
             ->children()
                 ->arrayNode('drivers')
-                    ->defaultValue([SyliusGridBundle::DRIVER_DOCTRINE_ORM])
```
diimpp, line 43: I think this will break a lot of end user setups as usually nobody specifies driver. If this true, it can only be merged in next major release.

src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -55,11 +55,29 @@ public function it_requires_only_grid_name(): void
     /**
      * @test
      */
-    public function it_uses_doctrine_orm_as_default_driver(): void
+    public function it_configured_no_default_drivers(): void
```
diimpp:
```suggestion
    public function it_configures_no_default_drivers(): void
```

src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -55,11 +55,29 @@ public function it_requires_only_grid_name(): void
...
+        $this->assertProcessedConfigurationEquals(
+            [
+                ['drivers' => [
+                    'doctrine/orm',
+                ]],
+            ],
+            [
+                'drivers' => [
+                    'doctrine/orm',
+                ],
```
diimpp, line 79: Everything is alright over here?

.github/workflows/build.yml
```diff
@@ -119,8 +119,44 @@ jobs:
...
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: "${{ matrix.php }}"
+                    coverage: none
+
+            -
+                name: Create-project with skeleton
+                run: |
+                    set -x
+                    php -v
```
diimpp:
```suggestion
```

.github/workflows/build.yml
```diff
@@ -119,8 +119,44 @@ jobs:
...
+        name: "Tests with the Symfony Skeleton PHP ${{ matrix.php }}${{ matrix.skeleton != '' && format(', Skeleton {0}', matrix.skeleton) || '' }}"
+        strategy:
+            fail-fast: false
+            matrix:
+                php: ["8.3"]
+                skeleton: ["^5", "^6", "^7"]
+
+        steps:
+            -
+                uses: actions/checkout@v2
```
diimpp:
```suggestion
                uses: actions/checkout@v4
```

There is another `v2 -> v4` at line 37

src/Bundle/DependencyInjection/Configuration.php
loic425: When we do not defined any driver, it enable all available ones.

src/Bundle/Tests/DependencyInjection/ConfigurationTest.php
loic425: yes, feel strange too, but that's ok.

src/Bundle/DependencyInjection/Configuration.php
diimpp: Hmm, yeah, I've mixed it with sylius/resource driver.

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -77,4 +75,97 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
...
+
+        return $availableDrivers;
+    }
+
+    /**
+     * @param string[] $configuredDrivers
+     * @param string[] $availableDrivers
+     * @param array<string, string> $gridDrivers
+     */
+    private function checkConfiguredDrivers(array $configuredDrivers, array $availableDrivers, array $gridDrivers): void
```
diimpp: In validation, `check` is usually reserved for function that returns `:bool` as result, while `assert` is for `:void` with exception.

```suggestion
    private function assertDriversValid(array $configuredDrivers, array $availableDrivers, array $gridDrivers): void
```

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -77,4 +75,97 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
...
+            $loader->load(sprintf('services/integrations/%s.xml', $enabledDriver));
+        }
+    }
+
+    /**
+     * @param array<string, array{driver?: array{name: string}}> $drivers
+     *
+     * @return array<string, string>
+     */
+    private function getGridDrivers(array $drivers): array
```
diimpp: `GridDriver` sounds like something else than regular `Driver` from $availableDrivers, which is as far as I understand is not true.

```suggestion
    private function resolveDriversUsedInGrids(array $grids): array
```

src/Bundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -77,4 +75,97 @@ public function getConfiguration(array $config, ContainerBuilder $container): Co
...
+    private function loadPersistence(array $drivers, array $grids, LoaderInterface $loader, ContainerBuilder $container): void
+    {
+        $availableDrivers = $this->getAvailableDrivers($container);
+
+        // Enable all available drivers if there is no configured drivers
+        $drivers = [] !== $drivers ? $drivers : $availableDrivers;
+
+        $gridDrivers = $this->getGridDrivers($grids);
+
+        $this->checkConfiguredDrivers($drivers, $availableDrivers, $gridDrivers);
```
diimpp: This code is a bit hard to understand, I've left previous comments to illustrate it, but here is a resume.

```php
getAvailableDrivers()
resolveDriversUsedInGrids()
assertDriversValid()
foreach ($drivers in $driver)
...
```

src/Bundle/Resources/config/services/integrations/twig/templating.xml
diimpp, line 1: A minor point of concern: While `Templating\Helper` is meant to be used by twig functions, it doesn't have any twig specific code or twig dependencies, but regardless will be enabled only with twig presence.

It might have been extracted from `Twig\Helper` to `Templating\Helper` so other possible templating engines could take advantage of it. 

I've no resume on what to do it with it -- as-is is alright.

diimpp (approved):

NoResponseMate (approved):

PR #377, opened 2025-05-20
Allow to define a template in action definition

loic425:
Replace #376 

```php
final class SubscriptionGrid extends AbstractGrid 
{
    public function buildGrid(GridBuilderInterface $gridBuilder): void
    {
        $gridBuilder
            ->addActionGroup(
                ItemActionGroup::create(
                    Action::create('show', 'custom_show')
                        ->setTemplate('grid/action/show.html.twig')
                    ,
                )
            )
        ;
    }
}           
``` 

⚠️ We cannot use this feature in the test application for now, cause the TwigGridRenderer from Resource bundle needs to use this template getter first.

In the package, there are two ways of defining the template.
1/ On Field => within options property
2/ On Filter => directly within template property

Indeed the grid could be used in non-Twig/non-HTML context. 
Definining the template in the options could be more detached to "HTML", but I think it's simpler for the DX to define this within a simple setTemplate. For now we use grids within HTML in 100% of the case, and even I already have some PoC to define console operations with grids, this is not bad to keep that getter/setter unused there.

### DIFF
--- a/src/Bundle/Builder/Action/Action.php
+++ b/src/Bundle/Builder/Action/Action.php
@@ -23,6 +23,8 @@ final class Action implements ActionInterface
 
     private ?bool $enabled = null;
 
+    private ?string $template = null;
+
     private ?string $icon = null;
 
     private array $options = [];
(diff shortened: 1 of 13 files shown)


diimpp (approved):
The long awaited feature :+1: It always felt weird to be forced to define new type just to use alternative template.

GSadee (approved):

GSadee:
Thank you @loic425! 🎉

PR #385, opened 2025-07-31
Moving the doctrine mappings from yaml to php attributes

### LINKED ISSUE #384: [GridBundle] Convert Doctrine mapping in attributes in the Test Application
It will fix that issue

Image

Related to #382

mamazu:
Closes #384


tests/Application/src/Entity/Author.php
```diff
@@ -15,32 +15,42 @@
...
+#[ORM\Table(name: 'app_author')]
 class Author implements ResourceInterface
 {
     /**
      * @Serializer\Expose
      *
      * @Serializer\Type("integer")
      */
+    #[ORM\Id]
+    #[ORM\Column(type: 'integer')]
```
mamazu, line 35: I've seen that in some versions there are Type classes for those doctrine columns. But I don't know if all versions that we support now have this feature.

mamazu:
What is the lowest version of the bundle that should have those changes? I can downport them.

tests/Application/src/Entity/Book.php
```diff
@@ -35,25 +38,41 @@ class Book implements ResourceInterface
...
     private string $state;
 
+    #[ORM\Column(type: 'boolean')]
     private bool $enabled = true;
 
+    #[ORM\Column(type: 'datetime')]
+    protected $createdAt = null;
+
+    #[ORM\Column(type: 'datetime', nullable: true)]
+    protected $updatedAt = null;
```
mamazu: Accoring to the yaml they were date_time columns but at least created at according to the constructor is a datetime_immutable. Should we fix that as well?

mamazu:
@loic425 Do you happen to have an idea. I think those errors have nothing to do with the change. But if you have a clue feel free to enlighten me.

loic425:
> @loic425 Do you happen to have an idea. I think those errors have nothing to do with the change. But if you have a clue feel free to enlighten me.

Hum, strange one but it seems to be linked a little. The grid maker is tested with an entity from the test application.

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -90,13 +90,13 @@ public function it_can_create_grids_interactively(): void
...
 
-        $tester->setInputs([Price::class]);
+        $tester->setInputs([AdminUser::class]);
         $tester->execute(['--namespace' => 'Tests\Tmp\Grid']);
 
-        $this->assertFileExists(self::tempFile(self::PRICE_GRID_PATH));
-        $this->assertSame(self::getPriceGridExpectedContent(), \file_get_contents(self::tempFile(self::PRICE_GRID_PATH)));
+        $this->assertFileExists(self::tempFile(self::ADMIN_USER_GRID_PATH));
+        $this->assertSame(self::getAdminUserGridExpectedContent(), \file_get_contents(self::tempFile(self::ADMIN_USER_GRID_PATH)));
     }
```
mamazu, line 100: Found the error, the `Price` class is not an entity (it's an embedable), and when switching to attributes this will no longer be returned as an entity. But this is not a bug in the test, so we'll just update them here.

tests/Application/src/Entity/Book.php
```diff
@@ -13,17 +13,19 @@
 
 namespace App\Entity;
 
+use DateTimeInterface;
```
loic425: not needed, use \DateTimeInterface instead.

loic425 (approved):

Issue #267, opened 2022-10-26
`setOptions` on `TwigField` will remove the template


mamazu:
## The problem

The `TwigField` is basically just a normal field with a default template. The problem is if you want to change the options of this field it will either have to specify the template again or you will lose it.

## Example

```php
<?php

$field = TwigField::create('some_field', '@Sylius/some_template.html.twig')
    ->setOptions(['vars' => ['labels' => 'path/to/label'])
;
```

Will result in the options being:
```
[
    'vars' => ['labels' => 'path/to/label']
]
```

and not:

```
[
    'template' => '@Sylius/some_template.html.twig',
    'vars' => ['labels' => 'path/to/label']
]
```

## Possible solution

The quick and easy answer would be to have the `TwigField` extend the `Field` and change the behavior of the `setOptions` function to merge with the template. This would be a big change though.

loic425:
@mamazu You're right. FYI, there is a "addOption". @lchrusciel merged too quickly the notice about this issue in the doc.
But I'm in favour of the merge options with the setOptions, that was my initial idea when implementing the grid builder. I think it'd be better for DX as we already have users who have that issue.

mamazu:
So extending it would be the best idea? I can make a merge request for that if you want.

mamazu:
I just had a look at this ticket again. I have two ideas for this. Either we deprecate the `setOptions` and forward people to the `addOptions` call which already does the merging or we keep the set options (as this can be quite helpful if you really want to remove / change a lot of options at once). However we then should also add a warning if the current options contain a twig field which would be unset by setting the options again. Then the user can decide if that's the right course of action or not.

loic425:
@mamazu On the Symfony OptionsResolver, the setDefaults merge current defaults with new ones.
I think it's more intuitive.
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/OptionsResolver/OptionsResolver.php#L277

But in this case, maybe we should deprecate addOptions instead.

mamazu:
Okay I'll do that.

PR #390, opened 2025-08-28
Deprecating setOptions and replacing it with addOptions

mamazu:
As described in #267 the setOptions will override any existing fields. You most of the time want to merge it which is what "addOptions" does.

### DIFF
--- a/src/Bundle/Builder/Field/DateTimeField.php
+++ b/src/Bundle/Builder/Field/DateTimeField.php
@@ -18,7 +18,7 @@ final class DateTimeField
     public static function create(string $name, string $format = 'Y-m-d H:i:s', ?string $timezone = null): FieldInterface
     {
         return Field::create($name, 'datetime')
-            ->addOptions(['format' => $format, 'timezone' => $timezone])
+            ->withOptions(['format' => $format, 'timezone' => $timezone])
         ;
     }
 }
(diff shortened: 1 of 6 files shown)


loic425:
Hum, we want to deprecate addOptions instead... so could we close this one and implement the merge in the setOptions?

mamazu:
Hm, I don't think that this is a good idea. When I set something I expect the value to be the value that I specified. Something like `mergeOptions` would probably be the best name and then we deprecate both.

loic425:
@mamazu What do you think about withOptions?
We'll talk about it for the 1.15 release, cause 1.14 is already stable.

For 1.15, we could work on these improvements and deprecations. I'd like to introduce the Grid live components on 1.15 too.

mamazu:
That sounds like a good idea. I just noticed that this "setOptions / addOptions" topic is also present in the Filters. Do we also want to update them to use the same methods?

GSadee (approved):

GSadee:
Thank you @mamazu! 🎉

loic425:
> That sounds like a good idea. I just noticed that this "setOptions / addOptions" topic is also present in the Filters. Do we also want to update them to use the same methods?

Yes, you could go with the other setter methods in 1.15 branch (I will create it)

PR #401, opened 2025-10-09
Convert phpspec 2 phpunit on component filter

loic425:


Prometee (approved):

src/Component/Tests/Unit/Filter/DateFilterTest.php
```diff
@@ -0,0 +1,237 @@
...
+
+        $expressionBuilder->expects($this->exactly(2))
+            ->method($this->logicalOr('greaterThanOrEqual', 'lessThan'))
+            ->willReturnCallback(function (string $field, string $value) {
+                return $value === '2016-12-05 08:00' ? 'EXPR1' : 'EXPR2';
+            });
+
+        $dataSource->expects($this->exactly(2))
+            ->method('restrict')
+            ->withConsecutive(['EXPR1'], ['EXPR2']);
```
Rafikooo: The method `withConsecutive()` is deprecated in PHPUnit 9 and removed in PHPUnit 10. Once we bump the dependency one day, it will break. I recommend to adjust it now so we don't have to iterate over this migration twice.

Refs: 
https://github.com/Sylius/SyliusGridBundle/actions/runs/19514483254/job/55862541450?pr=401#step:8:79
https://tomasvotruba.com/blog/how-to-upgrade-deprecated-phpunit-with-consecutive

loic425: it will need to be simplified after PHPUnit Bump, but here is the result:
```php
$dataSource->expects($invokedCount)
  ->method('restrict')
  ->willReturnCallback(function (string $expression) use ($invokedCount) {
      $numberOfInvocationMethod = method_exists($invokedCount, 'numberOfInvocations') ? 'numberOfInvocations' : 'getInvocationCount';
  
      if ($invokedCount->$numberOfInvocationMethod() === 1) {
          $this->assertEquals('EXPR1', $expression);
      }
  
      if ($invokedCount->$numberOfInvocationMethod() === 2) {
          $this->assertEquals('EXPR2', $expression);
      }
  });
```

Rafikooo (approved):

PR #402, opened 2025-10-10
Convert phpspec 2 phpunit on component filtering

loic425:


Prometee (commented):
Reviewing...

src/Component/spec/Filtering/FiltersCriteriaResolverSpec.php
```diff
@@ -11,78 +11,115 @@
...
-
-        $this->hasCriteria($grid, $criteriaParameters)->shouldReturn(true);
-
-        $grid->getFilters()->willReturn([$filter]);
-        $filter->getCriteria()->willReturn('czapla');
+        $this->assertInstanceOf(FiltersCriteriaResolverInterface::class, $this->filtersCriteriaResolver);
+    }
 
-        $this->hasCriteria($grid, $emptyParameters)->shouldReturn(true);
+    public function testChecksWhetherAnyCriteriaAreAvailable(): void
```
Prometee: There is a way this method can be refactored to use PHPUnit DataProvider method to simplify the code and each specific cases. WDYT?

Prometee (approved):

src/Component/spec/Filtering/FiltersCriteriaResolverSpec.php
loic425: I'm not very familiar with DataProvider system.  Thus, I've used chatGPT to create it, please review this carefully.

src/Component/Tests/Unit/Filtering/FiltersCriteriaResolverTest.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->filtersCriteriaResolver = new FiltersCriteriaResolver();
+    }
+
+    public function testItImplementsFiltersCriteriaResolverInterface(): void
+    {
+        $this->assertInstanceOf(FiltersCriteriaResolverInterface::class, $this->filtersCriteriaResolver);
+    }
+
+    /**
+     * @dataProvider criteriaDataProvider
```
Prometee, line 38: You can use PHP attributs instead to avoid changing it in a while when Annotations will be deprecated

Prometee (approved):

src/Component/Tests/Unit/Filtering/FiltersCriteriaResolverTest.php
loic425: Available in PHPUnit 10. We need to upgrade first. I'm sure there is Rector rule.

src/Component/Tests/Unit/Filtering/FiltersCriteriaResolverTest.php
```diff
@@ -0,0 +1,182 @@
...
+final class FiltersCriteriaResolverTest extends TestCase
+{
+    private FiltersCriteriaResolver $filtersCriteriaResolver;
+
+    protected function setUp(): void
+    {
+        $this->filtersCriteriaResolver = new FiltersCriteriaResolver();
+    }
+
+    public function testItImplementsFiltersCriteriaResolverInterface(): void
```
Rafikooo, line 32:
```suggestion
    public function testImplementsFiltersCriteriaResolverInterface(): void
```
For sake of consistency

Rafikooo (approved):

PR #399, opened 2025-10-09
Convert phpspec 2 phpunit on component field types

loic425:

PR #407, opened 2025-10-14
Improve enum support with EnumField and EnumFilter

acrobat:
I've added a EnumField type to render database enum values correctly as an Enum can't be converted to string (for the string type). In this field I've used the same logic as in the Symfony Enum form type, eg. if it's an translatable enum, translate it first otherwise just render the ->value of the enum

The same goes for the filter where we use the Symfony Enum form type to render the filter, the only config it needs is the enum class used to render the options.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -77,6 +77,7 @@
         "symfony/dotenv": "^6.4 || ^7.0",
         "symfony/maker-bundle": "^1.36",
         "symfony/polyfill-mbstring": "<1.22.0 || >1.22.0",
+        "symfony/translation": "^6.4 || ^7.0",
         "symfony/twig-bundle": "^6.4 || ^7.0",
         "twig/twig": "^2.12 || ^3.0",
         "vimeo/psalm": "^5.23",
--- a/src/Bundle/Resources/config/services/field_types.xml
+++ b/src/Bundle/Resources/config/services/field_types.xml
@@ -34,5 +34,12 @@
             <tag name="sylius.grid_field" type="string" />
         </service>
         <service id="Sylius\Component\Grid\FieldTypes\StringFieldType" alias="sylius.grid_field.string" />
+
+        <service id="sylius.grid_field.enum" class="Sylius\Component\Grid\FieldTypes\EnumFieldType">
+            <argument type="service" id="sylius.grid.data_extractor" />
+            <argument type="service" id="translator" on-invalid="null" />
+            <tag name="sylius.grid_field" type="enum" />
+        </service>
+        <service id="Sylius\Component\Grid\FieldTypes\EnumFieldType" alias="sylius.grid_field.enum" />
     </services>
 </container>
--- a/src/Component/FieldTypes/EnumFieldType.php
+++ b/src/Component/FieldTypes/EnumFieldType.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\FieldTypes;
+
+use Sylius\Component\Grid\DataExtractor\DataExtractorInterface;
+use Sylius\Component\Grid\Definition\Field;
+use Sylius\Component\Grid\Exception\LogicException;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Contracts\Translation\TranslatableInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+use Webmozart\Assert\Assert;
+
+final class EnumFieldType implements FieldTypeInterface
+{
+    public function __construct(
+        private DataExtractorInterface $dataExtractor,
+        private ?TranslatorInterface $translator = null,
+    ) {
+    }
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $enum = $this->dataExtractor->get($field, $data);
+
+        if (null === $enum) {
+            return '';
+        }
+
+        Assert::isInstanceOf($enum, \UnitEnum::class);
+
+        if ($enum instanceof TranslatableInterface) {
+            if (null === $this->translator) {
+                throw new LogicException('You have configured a translatable enum, but Symfony translator is not available. Try running "composer require symfony/translator".');
+            }
+
+            return $enum->trans($this->translator);
+        }
+
+        if ($enum instanceof \BackedEnum) {
+            return (string) $enum->value;
+        }
+
+        return $enum->name;
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefined('vars');
+        $resolver->setAllowedTypes('vars', 'array');
+    }
+}
--- a/src/Component/Tests/Unit/FieldTypes/EnumFieldTypeTest.php
+++ b/src/Component/Tests/Unit/FieldTypes/EnumFieldTypeTest.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Tests\Unit\FieldTypes;
+
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Sylius\Component\Grid\DataExtractor\DataExtractorInterface;
+use Sylius\Component\Grid\Definition\Field;
+use Sylius\Component\Grid\Exception\LogicException;
+use Sylius\Component\Grid\FieldTypes\EnumFieldType;
+use Sylius\Component\Grid\FieldTypes\FieldTypeInterface;
+use Sylius\Component\Grid\Tests\Dummy\TestEnum;
+use Sylius\Component\Grid\Tests\Dummy\TestNonBackedEnum;
+use Sylius\Component\Grid\Tests\Dummy\TestTranslatableEnum;
+use Symfony\Contracts\Translation\TranslatorInterface;
+use Webmozart\Assert\InvalidArgumentException;
+
+final class EnumFieldTypeTest extends TestCase
+{
+    private DataExtractorInterface|MockObject $dataExtractorMock;
+
+    private EnumFieldType $enumFieldType;
+
+    private Field $enumField;
+
+    protected function setUp(): void
+    {
+        $this->dataExtractorMock = $this->createMock(DataExtractorInterface::class);
+        $this->enumFieldType = new EnumFieldType($this->dataExtractorMock, $this->createMock(TranslatorInterface::class));
+        $this->enumField = Field::fromNameAndType('status', 'enum');
+    }
+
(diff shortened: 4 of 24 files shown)


src/Component/spec/FieldTypes/EnumFieldTypeSpec.php
```diff
@@ -0,0 +1,65 @@
...
+use Sylius\Component\Grid\DataExtractor\DataExtractorInterface;
+use Sylius\Component\Grid\Definition\Field;
+use Sylius\Component\Grid\FieldTypes\FieldTypeInterface;
+use Sylius\Component\Grid\Tests\Dummy\TestEnum;
+use Sylius\Component\Grid\Tests\Dummy\TestNonBackedEnum;
+use Sylius\Component\Grid\Tests\Dummy\TestTranslatableEnum;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+final class EnumFieldTypeSpec extends ObjectBehavior
```
loic425: Please, create a PHPUnit test instead.
We are currently removing all Phpspec tests.

composer.json
```diff
@@ -40,6 +40,7 @@
         "symfony/http-kernel": "^6.4 || ^7.0",
         "symfony/options-resolver": "^6.4 || ^7.0",
         "symfony/property-access": "^6.4 || ^7.0",
+        "symfony/translation": "^6.4 || ^7.0",
```
loic425: please move on dev requirements to keep it optional.

src/Bundle/Resources/config/services/field_types.xml
```diff
@@ -34,5 +34,12 @@
             <tag name="sylius.grid_field" type="string" />
         </service>
         <service id="Sylius\Component\Grid\FieldTypes\StringFieldType" alias="sylius.grid_field.string" />
+
+        <service id="sylius.grid_field.enum" class="Sylius\Component\Grid\FieldTypes\EnumFieldType">
+            <argument type="service" id="sylius.grid.data_extractor" />
+            <argument type="service" id="translator" />
```
loic425:
```suggestion
            <argument type="service" id="translator" on-invalid="null" />
```

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,53 @@
...
+final class EnumFieldType implements FieldTypeInterface
+{
+    private DataExtractorInterface $dataExtractor;
+    private TranslatorInterface $translator;
+
+    public function __construct(DataExtractorInterface $dataExtractor, TranslatorInterface $translator)
+    {
+        $this->dataExtractor = $dataExtractor;
+        $this->translator = $translator;
+    }
```
loic425:
```suggestion
    public function __construct(
        private DataExtractorInterface $dataExtractor, 
        private ?TranslatorInterface $translator = null,
   ) {
    }
```

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,53 @@
...
+
+        if (!$enum instanceof \BackedEnum) {
+            return $enum->name;
+        }
+
+        if (!$enum instanceof TranslatableInterface) {
+            return (string) $enum->value;
+        }
+
+        return $enum->trans($this->translator);
```
loic425:
```suggestion
if (null === $this->trans) {
            throw new LogicException('You have configured a translatable enum, but Symfony translator is not available. Try running "composer require symfony/translator".');
        }

        return $enum->trans($this->translator);
```

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,53 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\FieldTypes;
+
+use Sylius\Component\Grid\DataExtractor\DataExtractorInterface;
+use Sylius\Component\Grid\Definition\Field;
```
loic425, line 17:
```suggestion
use Sylius\Component\Grid\Definition\Field;
use Sylius\Component\Exception\LogicException;
```

src/Component/Filter/EnumFilter.php
```diff
@@ -0,0 +1,37 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Filter;
+
+use Sylius\Component\Grid\Data\DataSourceInterface;
+use Sylius\Component\Grid\Filtering\FilterInterface;
+
+final class EnumFilter implements FilterInterface
+{
+    public function apply(DataSourceInterface $dataSource, string $name, $data, array $options): void
```
loic425: this code is the same on as the SelectFilter, so I think we could use the SelectFilter instead.

loic425 (changes requested):

acrobat:
Thanks for the review @loic425! I've processed your feedback and converted the spec test to a phpunit test!

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,54 @@
...
+        if (!$enum instanceof \BackedEnum) {
+            return $enum->name;
+        }
+
+        if (!$enum instanceof TranslatableInterface) {
+            return (string) $enum->value;
+        }
+
+        if (null === $this->translator) {
+            throw new \LogicException('You have configured a translatable enum, but Symfony translator is not available. Try running "composer require symfony/translator".');
```
loic425: Please use LogicException from the Grid namespace.

src/Component/FieldTypes/EnumFieldType.php
loic425: You need to use this exception, not the php native one.

loic425 (changes requested):

src/Component/FieldTypes/EnumFieldType.php
acrobat: Oops, I missed that one. Fixed!

loic425:
A few things to add:
- ~~A test of the admin user grids on GridUiTest~~
- admin user grid configuration on config (both yaml and php).
    - yaml one https://github.com/Sylius/SyliusGridBundle/blob/1.14/tests/Application/config/sylius/grids.yaml
    - php one https://github.com/Sylius/SyliusGridBundle/tree/1.14/tests/Application/config/sylius/grids

tests/Application/config/sylius/grids.yaml
```diff
@@ -208,3 +208,30 @@ sylius_grid:
...
+                status:
+                    type: enum
+                    options:
+                        enum_class: App\Enum\AdminUserStatusEnum
+                        multiple: true
+            fields:
+                title:
+                    type: callable
+                    options:
+                        callable: "callable:strtoupper"
```
loic425:
```suggestion
                    type: string
```

loic425 (changes requested):

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,55 @@
...
+    public function render(Field $field, $data, array $options): string
+    {
+        $enum = $this->dataExtractor->get($field, $data);
+
+        if (!$enum instanceof \BackedEnum) {
+            return $enum->name;
+        }
+
+        if (!$enum instanceof TranslatableInterface) {
+            return (string) $enum->value;
```
loic425: I think we'll need a try catch here.

Like this one implemented by Florian on Callable (you'll need adapt the code, of course).
```
try {
    $value = (string) $value;
} catch (\Throwable $e) {
    throw new UnexpectedValueException(\sprintf(
        'Callable field (name "%s") returned value could not be converted to string: "%s".',
        $field->getName(),
        $e->getMessage(),
    ));
}
```

acrobat: Can't we safely assume a cast will work as the Backenum::$value property only allows int or string

> A Backed Enum may be backed by types of int or string, and a given enumeration supports only a single type at a time (that is, no union of int|string). 
https://www.php.net/manual/en/language.enumerations.backed.php

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,59 @@
...
+        private ?TranslatorInterface $translator = null,
+    ) {
+    }
+
+    public function render(Field $field, $data, array $options): string
+    {
+        $enum = $this->dataExtractor->get($field, $data);
+
+        if (!$enum instanceof \UnitEnum) {
+            return '';
```
loic425, line 37: Should we throw an exception instead?

src/Component/Tests/Unit/FieldTypes/EnumFieldTypeTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function testAGridFieldType(): void
+    {
+        $this->assertInstanceOf(FieldTypeInterface::class, $this->enumFieldType);
+    }
+
+    public function testUsesDataExtractorToObtainDataAndRendersIt(): void
+    {
+        /** @var Field|MockObject $fieldMock */
+        $fieldMock = $this->createMock(Field::class);
```
loic425: I think you can use the real instance instead.
EnumField::create...

acrobat: Sure, I now based it on the other unit tests from https://github.com/Sylius/SyliusGridBundle/pull/399/files#diff-57fa8d133cf70faa3b036d9b7d5a703885049c50865b1b6db255fd43b25031ddR47

But I've updated the tests to reflect this!

src/Component/FieldTypes/EnumFieldType.php
acrobat: I added this case for null values but I solved it in another way (similar to the datetime field type) and added tests for the extra cases (null and invalid values)

acrobat:
@loic425 I've addressed your latest comments!

src/Component/FieldTypes/EnumFieldType.php
```diff
@@ -0,0 +1,62 @@
...
+
+        if (!$enum instanceof TranslatableInterface) {
+            return (string) $enum->value;
+        }
+
+        if (null === $this->translator) {
+            throw new LogicException('You have configured a translatable enum, but Symfony translator is not available. Try running "composer require symfony/translator".');
+        }
+
+        return $enum->trans($this->translator);
```
Prometee: Maybe I'm wrong since I'm not using the Enum so often, but does a non backed enum can have `TranslatableInterface`, if yes then this service must check first for the `TranslatableInterface` then fallback to the `->name` or `->value`.

WDYT?

loic425: Chat GPT warns about that.

image

loic425: his suggestion
image

acrobat: It is indeed possible that a non backed enum is translatable, but I think in the cases of grids which fetches the data from the DB it will always be a backed enum. But to be fully feature complete I've switched the checks around to make that possible!

Prometee: @acrobat Don't forget Grid can be used with simple DTO out of any DB 😉

loic425 (approved):

Prometee (approved):

src/Component/FieldTypes/EnumFieldType.php
acrobat: Yeah true, that's now supported! 🎉

NoResponseMate (approved):

PR #430, opened 2025-10-28
Add missing service aliases

loic425:


loic425:
I'd prefer to wait for https://github.com/Sylius/SyliusGridBundle/pull/436 merge. Cause it will have conflicts and this one is easier to fix.

src/Bundle/Resources/config/services.php
```diff
@@ -88,6 +91,9 @@
 
     $services->alias(GridConfigurationSortingHandlerInterface::class, 'sylius.grid.configuration_sorting_handler');
 
+    $services->alias('sylius.grid.provider', 'sylius.grid.chain_provider');
+    $services->alias(GridProviderInterface::class, 'sylius.grid.provider');
```
Prometee, line 95: This can be moved to line 125 to regroup the definitions, right ?

Prometee (changes requested):

src/Bundle/Resources/config/services.php
loic425: When this is the main provider thing, I prefer to put it on the top to know which is the main one for the alias.

Prometee (approved):

diimpp (approved):

NoResponseMate (approved):

PR #433, opened 2025-10-29
[Chore] Bump package versions

loic425:
- Drop support of Symfony 7.0 and 7.1
- Drop support of PHP 8.1
- Remove Psalm
- Add Symfony 7.4 beta in CI


src/Component/composer.json
```diff
@@ -31,16 +31,16 @@
         "php": "^8.1",
```
Rafikooo: I believe it should be updated.

Rafikooo (commented):
Maybe a note about the drops in 1.15 should be added to the upgrade file?

GSadee (approved):
Note the changes in supported versions in the UPGRADE file for 1.15

src/Component/composer.json
loic425: You are right, I'll change that after lunching break

loic425: done

loic425:
> Note the changes in supported versions in the UPGRADE file for 1.15

done

GSadee (approved):

PR #437, opened 2025-11-03
PHPStan configuration enhancements with Symfony extension support

Prometee:
This pull request enhances the PHPStan configuration by adding Symfony-specific static analysis support, improving configuration structure, and adding missing generic type annotations to Form Type classes.

**Configuration updates:**

* Added `phpstan/phpstan-symfony` extension for better Symfony-specific static analysis capabilities, including improved understanding of Symfony Form components, dependency injection, and framework-specific patterns
* Added `phpstan/extension-installer` to automatically discover and load PHPStan extensions, eliminating the need for manual extension includes in the configuration file
* Configured explicit `paths` parameter to analyze only the `src` directory, making the analysis scope clear and explicit
* Set PHPStan analysis `level` to `max` for the most comprehensive static analysis coverage
* Simplified `excludePaths` by removing the `%currentWorkingDirectory%` prefix, making paths more concise and readable
* Removed manual extension includes (`phpstan-webmozart-assert`, `phpstan-phpunit`) from `phpstan.neon.dist` as they are now automatically loaded by `extension-installer`

**Code quality improvements:**

* Added missing `@extends AbstractType<mixed>` PHPDoc annotations to all Filter Type classes:
  - `BooleanFilterType`
  - `DateFilterType`
  - `EntityFilterType`
  - `EnumFilterType`
  - `ExistsFilterType`
  - `MoneyFilterType`
  - `NumericRangeFilterType`
  - `SelectFilterType`
  - `StringFilterType`
* These annotations help PHPStan understand the generic type inheritance from Symfony's `AbstractType`, improving type inference and reducing false positives

PR #438, opened 2025-11-03
Phpstan array doc enhancements

Prometee:
This PR is based on #437 (waiting for it to be merged)

Add array PHPDoc to components interface and fix related PHPStan issues

### DIFF
--- a/src/Bundle/Config/GridConfig.php
+++ b/src/Bundle/Config/GridConfig.php
@@ -37,8 +37,8 @@ final class GridConfig implements GridConfigInterface
             'grids' => [],
         ];
 
-        foreach ($this->grids as $gridBuilder) {
-            $output['grids'][] = $gridBuilder->toArray();
+        foreach ($this->grids as $name => $gridBuilder) {
+            $output['grids'][$name] = $gridBuilder->toArray();
         }
 
         return $output;
--- a/src/Bundle/Config/GridConfig.php
+++ b/src/Bundle/Config/GridConfig.php
@@ -17,6 +17,9 @@ use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
 
 final class GridConfig implements GridConfigInterface
 {
+    /**
+     * @var array<string, GridBuilderInterface>
+     */
     private array $grids = [];
 
     public function addGrid(GridBuilderInterface $gridBuilder): GridConfigInterface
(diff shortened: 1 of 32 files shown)


src/Bundle/Config/GridConfig.php
```diff
@@ -28,10 +31,16 @@ public function addGrid(GridBuilderInterface $gridBuilder): GridConfigInterface
...
     public function toArray(): array
     {
-        $output = [];
+        if (count($this->grids) <= 0) {
+            return [];
+        }
+
+        $output = [
+            'grids' => [],
+        ];
```
diimpp, line 38: Is this intentional change, that previous empty result was returning `[]`, but now it will return `['grids' => []]` instead?

```suggestion
        $output = [];
```

src/Bundle/Config/GridConfigInterface.php
```diff
@@ -32,6 +32,8 @@ public function addGrid(GridBuilderInterface $gridBuilder): self;
 
         /**
          * Gets all configuration represented as an array.
+         *
+         * @return array<string, mixed>
```
diimpp, line 36: Is it necessary to provide more detailed shape since `grids` key is always present in current implementation? (<string, <int, mixed>> or something)

src/Component/Data/ExpressionBuilderInterface.php
```diff
@@ -79,11 +79,13 @@ public function greaterThan(string $field, $value);
     public function greaterThanOrEqual(string $field, $value);
 
     /**
+     * @param mixed[] $values
```
diimpp, line 82: For `in` and `notIn` at `src/Bundle/Doctrine/DBAL/ExpressionBuilder.php`  `$values` are configured to be array of strings earlier in this PR.

Prometee: The interface define a generic type, the classes implementing it have stricter types.

- DBAL implementation only accept `string[]`.
- ORM implementation accept `mixed`.
- PHPCRODM implementation accept `mixed[]`.

src/Bundle/Config/GridConfig.php
Prometee: Maybe you did not notice I flip the first if condition to make the method return earlier.

But the behavior of this method remain the same.

Prometee: @diimpp your comment makes me fix the CI, I forgot to set the array key.

src/Bundle/Config/GridConfigInterface.php
Prometee: Since it's an interface, I would let the implementation choose what we can find inside.
Also notice that this class is not yet documentated as deprected but it seems it will be (ping @loic425 ?)

diimpp (approved):

PR #447, opened 2025-11-17
[Chore] Remove resource bundle dependency on test application

loic425:
Partially #455

### 🔧 Context

The Grid bundle is already fully decoupled from SyliusResourceBundle. However, the test application used in this repository still included the Resource bundle, which was unnecessary and could lead to confusion about the actual dependencies of the Grid component.
This PR removes that extra layer so the test suite can better reflect the true independence of the Grid bundle.

### ✅ What this PR does

Removes SyliusResourceBundle from the test application, ensuring the Grid bundle is tested in an environment that matches its real requirements.

Improves the accuracy and reliability of the test suite by avoiding implicit dependencies coming from Resource.

**Helps prepare the Grid bundle for an independent release cycle**, since the Grid does not rely on Resource — while Resource does need to be tested with Grid.

### 🔄 Why this matters
- Clarifies the separation between Grid and Resource, ensuring the Grid bundle stands on its own.
- Simplifies the test setup and reduces the surface area for side-effects caused by other Sylius packages.
- Enables more consistent and predictable versioning: Grid evolves independently, while Resource continues to rely on it.
- Helps maintain a clean architecture across Sylius packages.


tests/Application/config/sylius/grids/author_with_books_with_fetch_join_collection_disabled.php
```diff
@@ -18,7 +18,7 @@
     $grid->addGrid(
         GridBuilder::create('app_author_with_books_with_fetch_join_collection_disabled')
         ->extends('app_author')
-        ->setRepositoryMethod(["expr:service('app.authors_with_books_query_builder')", 'create'])
+        ->setRepositoryMethod('createWithBooksQueryBuilder')
```
loic425, line 21: `expr` is only parsed by `Sylius\Bundle\ResourceBundle\Grid\Parser\OptionsParser` from the resource bundle, so we are not able to use it here without the resource package.

tests/Application/src/Controller/GridTrait.php
```diff
@@ -0,0 +1,83 @@
...
+use Sylius\Component\Grid\Parameters;
+use Sylius\Component\Grid\Provider\GridProviderInterface;
+use Sylius\Component\Grid\View\GridViewFactoryInterface;
+use Sylius\Component\Grid\View\GridViewInterface;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Contracts\Service\Attribute\Required;
+
+trait GridTrait
```
loic425, line 25: This trait is only for the test application. So we'll improve it later, cause we'll need to create a service at some point.

tests/Application/src/Controller/GridTrait.php
```diff
@@ -0,0 +1,83 @@
...
+    #[Required]
+    public function setRequestStack(
+        RequestStack $requestStack,
+    ): void {
+        $this->request = $requestStack->getCurrentRequest() ?? new Request();
+    }
+
+    #[Required]
+    public function setGridProvider(
+        #[Autowire(service: 'sylius.grid.chain_provider')]
```
loic425: An alias has been added recently (but not merged), we'll be able to remove this.

src/Bundle/Tests/Functional/Maker/MakeGridTest.php
```diff
@@ -353,7 +353,7 @@ private static function getBoardGameGridExpectedContent(): string
 
 namespace App\Tests\Tmp\Grid;
 
-use App\BoardGameBlog\Infrastructure\Sylius\Resource\BoardGameResource;
+use App\BoardGameBlog\Domain\Model\BoardGame;
```
loic425, line 356: This is not a Sylius resource anymore, but just a simple PHP class.
We still need this one to test the grid maker on non-Doctrine object.

tests/Application/src/Entity/Author.php
```diff
@@ -13,38 +13,25 @@
...
-use Sylius\Component\Resource\Model\ResourceInterface;
+use Doctrine\ORM\Mapping\Entity;
 
-/**
- * @Serializer\ExclusionPolicy("all")
- */
 #[ORM\MappedSuperclass]
 #[ORM\Table(name: 'app_author')]
-class Author implements ResourceInterface
+#[Entity(repositoryClass: AuthorRepository::class)]
```
NoResponseMate:
```suggestion
#[ORM\Entity(repositoryClass: AuthorRepository::class)]
```
The `Mapping` import is already in place, and it's already like that in the other entities 💃

loic425: done, you are right

NoResponseMate (approved):

PR #452, opened 2025-11-24
Remove ApiTestCase and use Foundry instead of Nelmio Alice

loic425:
Partially #455

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,9 +129,11 @@ jobs:
 
             -
                 name: Run lint container without Doctrine
+                if: false # disable it for now
                 run: |
-                    composer remove --dev doctrine/*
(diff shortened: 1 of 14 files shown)


.github/workflows/build.yml
```diff
@@ -133,9 +133,11 @@ jobs:
 
             -
                 name: Run lint container without Doctrine
+                if: false # disable it for now
```
loic425, line 132: Sylius Resource configuration requires the doctrine/orm driver to be available.
Before this PR, the Doctrine bundle was not removed cause api-test-case package requires it. 

We'll rework this part after sylius resource package removal from dependencies here (#447).

diimpp (approved):

NoResponseMate (approved):

PR #456, opened 2025-11-25
Remove Psalm :tada:

loic425:


Prometee (changes requested):
There is also multiple `@psalm` phpdoc to remove

Prometee (approved):

loic425:
> There is also multiple `@psalm` phpdoc to remove

We decided to remove these phpdoc in 1.15 instead.

Rafikooo (approved):

PR #457, opened 2025-11-25
Phpstan baseline final cleanup

Prometee:
This PR has already been approved in an other fork:

- https://github.com/Prometee/SyliusGridBundle/pull/2
- https://github.com/Prometee/SyliusGridBundle/pull/1

The second one has been merged into the first one and now this PR is regrouping both of them


src/Bundle/Builder/Action/Action.php
```diff
@@ -80,6 +81,9 @@ public function setIcon(string $icon): ActionInterface
         return $this;
     }
 
+    /**
+     * @param array<string, mixed> $options
```
diimpp: Duplicates `setOptions` at `ActionInterface`. Is it necessary?

loic425 (approved):

src/Component/Filter/DateFilter.php
```diff
@@ -41,11 +41,14 @@ public function apply(DataSourceInterface $dataSource, string $name, $data, arra
     {
         $expressionBuilder = $dataSource->getExpressionBuilder();
 
-        $field = (string) $this->getOption($options, 'field', $name);
+        /** @var string $field */
+        $field = $options['field'] ?? $name;
```
diimpp, line 48: Do you still need to cast it to string, like it was before?

```suggestion
        $field = $options['field'] ? (string) $options['field'] : $name;
```

or the way how it's done bellow

```php
            /** @var bool|string|int $inclusive */
            $inclusive = $options['inclusive_from'] ?? self::DEFAULT_INCLUSIVE_FROM;
            $inclusive = (bool) $inclusive;
```

Prometee: I simplify the PHPDoc by defining the array shape on the method.
It allows to really understand what is expected here.

NoResponseMate (approved):

PR #461, opened 2025-11-28
feat(Metadata): Invokable grids

loic425:


src/Bundle/Grid/AbstractGrid.php
```diff
@@ -47,7 +41,23 @@ public function toArray(): array
             throw new LogicException(sprintf('The configured build method "%s" does not exist.', $buildMethod));
         }
 
-        $this->$buildMethod($gridBuilder);
+        $grid = [$this, $buildMethod];
+        Assert::true(is_callable($grid));
```
mamazu: A helpful error message would be nice here as well.

loic425: This is a PoC. I'm not satisfied yet :)

PR #466, opened 2025-12-03
Fix 1.15 branch CI

loic425:


tests/Application/config/packages/sylius_grid.yaml
```diff
@@ -1,13 +1,5 @@
 sylius_grid:
     templates:
-        action:
```
loic425, line 3: none of these templates are ok without using the resource bundle.

NoResponseMate (approved):

PR #480, opened 2026-06-16
feat(DX): Invokable grids

loic425:
It replaces #461

Features:
- No more grid building logic inside the grid (the GridBuilder was created into the AbstractGrid)
- We can use any php grid with a simple "sylius.grid" tag 

**Before**
```php
<?php

declare(strict_types=1);

namespace App\Grid;

use App\Entity\Book;
use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
use Sylius\Bundle\GridBundle\Grid\AbstractGrid;
use Sylius\Component\Grid\Attribute\AsGrid;

#[AsGrid(resourceClass: Book::class, name: 'app_book')]
final class BookGrid extends AbstractGrid
{
    public function __invoke(GridBuilderInterface $gridBuilder): void
    {
    }
}

```

**After**
```php
<?php

declare(strict_types=1);

namespace App\Grid;

use App\Entity\Book;
use Sylius\Bundle\GridBundle\Builder\GridBuilderInterface;
use Sylius\Component\Grid\Attribute\AsGrid;

#[AsGrid(resourceClass: Book::class, name: 'app_book')]
final class BookGrid
{
    public function __invoke(GridBuilderInterface $gridBuilder): void
    {
    }
}

```

TODO:
- [x] Deprecate the AbstractGrid class
- [x] Deprecate the Sylius\Bundle\GridBundle\Grid\GridInterface

PR #483, opened 2026-06-19
Remove unnecessary configuration

loic425:
Related to #https://github.com/Sylius/SyliusGridBundle/pull/480#discussion_r3440224395


src/Bundle/DependencyInjection/Compiler/AttributeGridPass.php
```diff
@@ -59,16 +59,9 @@ public function process(ContainerBuilder $container): void
                 /** @var string $class */
                 $class = $attribute['class'] ?? throw new LogicException(sprintf('"class" attribute not found on "sylius.grid" tag for "%s" service.', $id));
 
-                /** @var string $name */
```
loic425, line 62: This is not necessary cause we have
`/** @var array<string, string|null> $attribute */`

GSadee (approved):

PR #487, opened 2026-06-25
feat(php-grids): Grid mutators

loic425:
Example:
```php
<?php

declare(strict_types=1);

namespace App\Grid;

use Sylius\Component\Grid\Attribute\AsGridMutator;
use Sylius\Component\Grid\Mutator\GridMutatorInterface;

#[AsGridMutator(grid: 'sylius_admin_product')]
final class OrderByNameProductGridMutator implements GridMutatorInterface
{
    public function __invoke(GridBuilderInterface $gridBuilder): void
    {
        $gridBuilder->orderBy('name', 'desc');
        $gridBuilder->removeField('code');
    }
}
```

With a priority:

```php
<?php

declare(strict_types=1);

namespace App\Grid;

use Sylius\Component\Grid\Attribute\AsGridMutator;
use Sylius\Component\Grid\Mutator\GridMutatorInterface;

#[AsGridMutator(grid: 'sylius_admin_product', priority: 20)]
final class RemoveImageProductGridMutator implements GridMutatorInterface
{
    public function __invoke(GridBuilderInterface $gridBuilder): void
    {
        $gridBuilder->removeField('image');
    }
}
```

### DIFF
--- a/src/Bundle/DependencyInjection/Compiler/GridMutatorPass.php
+++ b/src/Bundle/DependencyInjection/Compiler/GridMutatorPass.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\GridBundle\DependencyInjection\Compiler;
+
+use Sylius\Component\Grid\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class GridMutatorPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if (!$container->hasDefinition('sylius.grid.mutator_collection')) {
+            return;
+        }
+
+        $definition = $container->getDefinition('sylius.grid.mutator_collection');
+
+        /** @var list<array{id: string, grid: string, priority: int}> $mutators */
+        $mutators = [];
+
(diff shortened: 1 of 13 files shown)


mamazu (approved):

src/Bundle/DependencyInjection/Compiler/GridMutatorPass.php
```diff
@@ -0,0 +1,48 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\GridBundle\DependencyInjection\Compiler;
+
+use Sylius\Component\Grid\Exception\LogicException;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+final class GridMutatorPass implements CompilerPassInterface
```
mamazu, line 21: Why don't we use the tagged_iterator for this?

src/Component/Attribute/AsGridMutator.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Grid\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
```
NoResponseMate: How about making it repeatable?

loic425: Hum why not, I need to adapt the compiler pass, but ok.

src/Bundle/DependencyInjection/Compiler/GridMutatorPass.php
loic425: I do not remember exactly. I did like this on Sylius Resource and APIP metadata mutators. I think the point is there are several mutators for the same grid. So grid code is not a unique key.

src/Component/Attribute/AsGridMutator.php
loic425: Done :)

NoResponseMate (approved):

PR #488, opened 2026-07-01
feat(mutators): Add support for mutators in legacy grids

loic425:


src/Bundle/Registry/MockGridRegistry.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Sylius\Bundle\GridBundle\Registry;
+
+use Sylius\Bundle\GridBundle\Grid\GridInterface as LegacyGridInterface;
+use Sylius\Component\Grid\GridInterface;
+
+/**
+ * @experimental
+ */
+final class MockGridRegistry implements GridRegistryInterface
```
loic425: That could be ArrayGridRegistry, as we use the ArrayLoader on the Symfony translator for PHPUnit tests.

loic425: I've replaced with ArrayGridRegistry, I think this is better. It could be used instead of the current GridRegistry which uses a ServiceLocator at some point.

loic425: I'm trying sth else...

loic425: Ok, that works this way:
```php
$locatedGrids = [];
foreach ($grids as $grid) {
    $locatedGrids[$grid->getName()] = fn () => $grid;
}

$gridRegistry = new GridRegistry(new ServiceLocator($locatedGrids));
```

So we don't need any other grid registry.

src/Bundle/Provider/ServiceGridProvider.php
```diff
@@ -110,7 +110,16 @@ private function extend(array $gridConfiguration, string $parentGridCode): array
...
         if ($grid instanceof LegacyGridInterface) {
-            return $grid->toArray();
+            $gridBuilder = GridBuilder::create($grid->getName());
+
+            foreach ($this->gridMutatorCollection->get($grid->getName()) as $mutator) {
+                ($mutator)($gridBuilder);
+            }
+
+            /** @var array<string, mixed> $newGridConfiguration */
+            $newGridConfiguration = array_replace_recursive($grid->toArray(), $gridBuilder->toArray());
```
TheMilek: Here I'm not sure if this doesn't force default driver in configuration by overwriting it silently 🤔 https://github.com/Sylius/SyliusGridBundle/blob/1.16/src/Bundle/Builder/GridBuilder.php#L24

loic425: Done, I've updated the PR to set driver and driver options from the current grid before calling the mutators.

TheMilek (approved):

PR #491, opened 2026-07-07
feat(deprecations): Deprecate the legacy grids and the grid events

loic425:


GSadee (approved):
Could you add also a note to the UPGRADE file about these deprecations?

PR #495, opened 2026-07-09
fix(interfaces): Move builder interfaces into the Grid component

loic425:
The Issue is here
```php
namespace Sylius\Component\Grid\Mutator;

interface GridMutatorInterface
{
    public function __invoke(GridBuilderInterface $gridBuilder): void;
}
```

The mutator interface is in the **component** and it uses a builder interface that is in the **bundle**.
But we also use the mutators in the ArrayProvider which is also in the component, so we cannot move the mutators to the bundle.

The easiest way is to move all the builder interfaces in the component. That what's this PR is doing.

To test if the interface aliases are ok, I've kept the legacy GridBuilderInterface usage in the BookGrid on the tests application.


GSadee (approved):
Add a proper note to the UPGRADE file about these changes and deprecations, I’m also considering adding `trigger_deprecation` to the bundle aliases, wdyt? 🤔

PR #500, opened 2026-09-23
fix(tests): Fix MakeGridTest failures and modernize PHPUnit metadata

loic425:
Generate ECS-compliant code from the grid maker skeleton: emit DateTimeField/TwigField imports only when used, sort AsGrid before GridBuilderInterface, and keep the class opening brace on its own line.

Also remove the unnecessary restore_exception_handler() calls that made tests risky, and replace doc-comment metadata (@test, @before, @dataProvider) with PHPUnit attributes to fix 91 PHPUnit 11 deprecations.

Related to
https://github.com/Sylius/SyliusGridBundle/actions/runs/35861043781/job/107181055013?pr=472


.github/workflows/build.yml
```diff
@@ -35,7 +35,7 @@ jobs:
                         symfony: "7.4.*"
                     -
                         php: 8.4
-                        symfony: "8.0.*"
```
Prometee: Is it possible to support any newer version using "^8.1" annotation ?

src/Bundle/Tests/Functional/Maker/MakerTestCase.php
```diff
@@ -13,14 +13,13 @@
...
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
 use Symfony\Component\Filesystem\Filesystem;
 
 abstract class MakerTestCase extends KernelTestCase
 {
-    /**
-     * @before
-     */
+    #[Before]
     public static function cleanupTmpDir(): void
```
Prometee: Could we make `cleanupTmpDir()` a non-static lifecycle method?

`#[Before]` is intended for instance methods, but this method remains declared as `public static`. As a result, PHPUnit may not execute it as expected before each test, leaving generated files in the temporary directory and making the tests order-dependent.

Could you change it to something like:

```suggestion
    #[Before]
    protected function cleanupTmpDir(): void
```

Alternatively, please use the appropriate PHPUnit lifecycle mechanism for a static setup method.

src/Bundle/Resources/config/skeleton/Grid.tpl.php
```diff
@@ -12,18 +28,21 @@
...
 use Sylius\Bundle\GridBundle\Builder\Action\ShowAction;
 use Sylius\Bundle\GridBundle\Builder\Action\UpdateAction;
+<?php if ($hasDateTimeFields): ?>
 use Sylius\Bundle\GridBundle\Builder\Field\DateTimeField;
+<?php endif; ?>
 use Sylius\Bundle\GridBundle\Builder\Field\StringField;
+<?php if ($hasBooleanFields): ?>
 use Sylius\Bundle\GridBundle\Builder\Field\TwigField;
-use Sylius\Component\Grid\Builder\GridBuilderInterface;
+<?php endif; ?>
```
Prometee: Could we avoid generating this import while the corresponding `TwigField::create()` call remains commented out?

Could you either remove this conditional import for now, or generate the `TwigField` call as actual code when a boolean field is detected?

src/Bundle/Resources/config/skeleton/Grid.tpl.php
Prometee: Could we normalize the field type here before performing the rendering checks?

The earlier loop normalizes `$type` with `strtoupper()` when determining which imports are needed, but this loop still compares the original value. If `$defaultFields` contains a lower- or mixed-case type, the template can generate an import without generating the corresponding field, or fail to generate the field despite detecting the type earlier.

Please normalize the type once and reuse the normalized value consistently for both import detection and field generation, for example:

```suggestion
                foreach ($defaultFields as $fieldname => $type) {
                    $type = strtoupper((string) $type);
```

Prometee (changes requested):

Prometee:
Great work on this PR, @loic425 ! The migration to PHPUnit attributes is clear and consistent, and the updates to the `make:grid` skeleton noticeably improve the quality of the generated code. Thanks for addressing the PHPUnit 11 deprecations and ECS compliance issues!

src/Bundle/Resources/config/skeleton/Grid.tpl.php
loic425: I've created this issue #501

.github/workflows/build.yml
loic425: We often need to deal with new deprecations, so I prefer to keep this like this, and support the newer version when needed.

loic425: And we don't have the "prefer-lowest" in this package CI.

Prometee (approved):

